PR #30, opened 2014-12-23
Add exit command

dgelessus:
`exit` returns with the numerical status it is given as an argument, useful for shell scripts to set their own exit status or to override that of the previous command.

(Also adds aliases `help` (== `man`) and `logout` (prints a helpful message), and safer eval of tuples from config.)


ywangd:
It does seem **not** possible to programmatically close a panel UI ... 
I'd still like to get some advices from @jsbain on this matter. Sorry for the extended delay. I'll give it a few more days then I'll merge the request. Thanks!

jsbain:
I've never been able to figure out how to close a panel.  
You can't re-present as full screen, `del` the view, use `View.close()`, ` View.hidden`, etc.

You could always present a new empty view in a panel, but I guess I don't see the use case.

Not sure if it is supported right now, but one could imagine running multiple Stash instances as a subview of some other view on ipad... Say, a tabbed view, or a window style system.   For instance you might be running ssh in one window/tab, and working locally in another.

In that case, `exit` at the top shell of an instance would close the tab or window...so you should still perhaps call close() to allow the possibility of non panel displays.

ywangd:
@jsbain allow multiple Stash instances is a great idea. It is not currently supported (different sessions have to be isolated from each other). it also involves major UI changes (based on my experience on UI coding). So I am not sure how much efforts it requires. But it is definitely something that I will explore in future updates. Thanks!

ywangd:
@dgelessus Thanks for the pull request.
I think you are right about Linux `exit` and `logout` commands. But for stash, it seems a bit verbose to have two commands for basically the same thing (stash does not have login). So I'll probably unify exit and logout commands into just exit in future updates. But before that, I'd like to work on some improvement on runtime. Hope it is OK with you. Thanks!

PR #38, opened 2014-12-30
inital tar.py

briarfox:
Added tar. It follows linux tar usage. 
Example extracts a gzip tar:
tar -vxzf test.tar.gz [some_directory]

Creates a bz2 archive:
tar -vcjf test.tar.bz2 dir1 file1.py file2.py

### DIFF
--- a/bin/tar.py
+++ b/bin/tar.py
@@ -1,0 +1,114 @@
+'''
+Used used with tar, gzip, bz2 files.
+
+Examples:
+    Create a gzip compressed archive:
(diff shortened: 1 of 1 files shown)


ywangd:
@briarfox Congratulations! Parenthood is a time consuming but very rewarding task. Hope you enjoy it.

Thanks for the pull request. It is definitely one of the fundamental commands for future features, e.g. package management. It should be safe to just let StaSh handle exceptions unless a command wants to print a customised error message or recover from a specific exception.

PR #45, opened 2015-01-04
pip

briarfox:
Here is an early version of pip. The formating is very bad, pythonista is crashing on backspace :(. I'll fix it soon. It's still missing features such as remove and update. It will work with any package archive that is .zip, .bz2, .gz. It does track installed packages and I'm hoping to parse previously installed packages and add them to the list.

### DIFF
--- a/bin/pip.py
+++ b/bin/pip.py
@@ -1,0 +1,198 @@
+'''
+serach - Searches for packages
+versions - Lists all versions for a package
+install - Installs the package off pypi
+list - Lists currently installed backages.
+
+usage: pip.py [-h] [-n RESULT_COUNT]
+              {search,versions,install,list} [package] [version]
+
+positional arguments:
+  {search,versions,install,list}
+  package
+  version
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -n RESULT_COUNT
+'''
+import xmlrpclib
+from ConfigParser import SafeConfigParser
+import argparse
+import tempfile
+import re
+import os
+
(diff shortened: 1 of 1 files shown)


ywangd:
@briarfox Thanks for working on `pip`. It looks pretty good already. I successfully installed `pyte` using it. I am not sure what do you mean by `pythonista is crashing on backspace`. I haven't experienced any crash yet. A few observations:
* `pip search pyte` does not return anything that is really `pyte`, while `pip versions pyte` does return all versions available for `pyte`
* pip seems to always try move a file named as `MODULE.py`, e.g. `pyte.py` from the `tmp_folder` to actual package folder. pyte does not have such a file and an error is reported. It however does NOT affect the overall success of the installation. Is this behaviour intended?

I personally wouldn't worry much about previously installed packages that are not managed by `pip`. Those packages may not even be available from pypi as they could be any 3rd party libraries.

briarfox:
@ywangd pythonista was crashing really bad for me. Omz's latest beta appears to be a fix for it. Pip search is returning the response from Pypi. It's pretty liberal, I'll see what I can do. I was being lazy and attempting to cover packages that had a folder and those that were a single script. I'll clean that up.

ywangd:
@briarfox That is interesting. I didn't have any crashes. Could you please give me a detailed example on how it crashed for you, like what package you were trying to install. I'll see if I can replicate it. 
If the search results is totally up to pypi, don't worry about it.

dgelessus:
PyPI search is confusing to say the least. If you're lucky the first result matches the search query, everything below that seems relatively random.

PR #46, opened 2015-01-04
Fixed mv added remove, update

briarfox:
Fixed the needless move. Added update and remove of pip tracked modules. The package info has been tweaked since last commit so you may need to remove modules to allow pip to track correctly.

I'm debating on adding an update all for updating of all pip installed packages. This could be a lengthy process and the user may want some modules to remain at a earlier version such as Twisted. Twisted 13.x.x works in pythonista where version 14.x.x does not.

### DIFF
--- a/bin/pip.py
+++ b/bin/pip.py
@@ -3,6 +3,8 @@ serach - Searches for packages
 versions - Lists all versions for a package
 install - Installs the package off pypi
 list - Lists currently installed backages.
+remove - remove a package installed by pip
+update - update a package installed by pip
 
 usage: pip.py [-h] [-n RESULT_COUNT]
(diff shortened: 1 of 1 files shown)


jsbain:
Just a thought: would it be possible to add a pip command which installs from a generic url?  Use case: install packages which are not hosted on pypi, such as on github.  For example the forked versions of dulwich and Gittle.  I'd like to use the package management/install/tracking capabilities of the pip command, but use external urls that point to zip or tar.gz files.

briarfox:
@jsbain Thats an interesting idea. Should be easy to implement. How do you see dealing with conflicts such as dulwich from pypi vs custom dulwich off github?

Issue #51, opened 2015-01-06
$? should not be listed by env command


ywangd:
The special variable `$?` should not be listed by the `env` command.

dgelessus:
I assume this also includes all the other special vars that a shell could create: $, !, #, *, @ and the numbers 0-9.

dgelessus:
Pull request @ #52.

ywangd:
Fixed with #52

PR #52, opened 2015-01-07
Exclude special envvars from list, as discussed in #51

dgelessus:


ywangd:
@dgelessus 
Thanks for the pull request. 
One comment: As `$N`, e.g. `$1`, is used for representing arguments to a shell script. It is possible that number of arguments is greater than 10. In this case, `$10`, `$11`, etc would also be created as special variables. On a command line, these variables have to be quoted like `${10}`. But in `os.environ`, they are just saved as `10`.

dgelessus:
I'm not entirely sure if that is the case. A number of sources indicate that only $1 through $9 are used for positional parameters and anything above that would require use of `$@`, `$*` or `shift`, though for example the GNU Bash reference specifically allows use of one or more digits (though numbers with multiple digits need to use brace syntax).

ywangd:
@dgelessus I didn't know that canonical Unix shell (sh) does not allow `$10` and up. But Bash for modern Linux including OS X do honour `$10` etc. So I think it is adequate for stash to support them as well. 

Check the first character of a variable like the follows
```python
vardict = {k: v for k, v in vardict if k[0] not in "$@?!#*0123456789"}
```
should be sufficient as it is impossible to enter a variable like `1A` from the command line.

ywangd:
@dgelessus If above changes are OK with you, I'll merge the pull and make the changes. Or if you would make the changes, I'll just simply merge afterwards. Thanks!

dgelessus:
Go ahead. I was thinking of adding a more complex check so names like ?foo wouln't be hidden, but that isn't worth the readability.

ywangd:
I have also made another correction to `vardict`, so the code now reads `... k, v in verdict.items() ...`. Thanks!

dgelessus:
Oops, I overlooked that. Thanks for fixing.

PR #68, opened 2015-01-17
md5sum added

briarfox:
I needed to get a md5 hash of a file so I added md5sum to stash. You can provide a filename and get the hash value or you can provide a txt file with md5 hashes and filenames to get a pass/fail on each file.

I'd like for md5sum to use pipes to allow text to be passed to it but I am unsure how it works.

### DIFF
--- a/bin/md5sum.py
+++ b/bin/md5sum.py
@@ -1,14 +1,15 @@
 '''
-md5sum - Get md5 hash of a file. Supports pipes.
+md5sum - Get md5 hash of a file or string. 
 
-usage: md5sum.py [-h] [-c] file
+usage: md5sum.py [-h] [-c] [file [file ...]]
 
 positional arguments:
-  file         File to get md5sum or file contailing a list of hashes and files
+  file         String or file to hash.
 
 optional arguments:
   -h, --help   show this help message and exit
-  -c, --check  Check a file with md5 values and files for a match. format:
+  -c, --check  Check a file with md5 hashes and file names for a match.
+               format:
                md5_hash filename
                md5_hash filename
                etc.
@@ -18,7 +19,9 @@ from Crypto.Hash import MD5
 import re
 import sys
 import os
-def get_md5(fileobj):
+import cStringIO
+
+def get_hash(fileobj):
     h = MD5.new()
     chunk_size = 8192
     while True:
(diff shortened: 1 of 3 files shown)


ywangd:
@briarfox  Thanks for the pull request.

In principle if a script can read from stdin, it can also read from a pipe. So `md5sum` has to be able to read form stdin, which it currently does not. 

A possible implementation is to allow the `file` positional argument to be optional and read from stdin if it does not present:

```Python
...
ap.add_argument('file',nargs='?',action='store',help='File to get md5sum or file contailing a list of hashes and files')
args = ap.parse_args(sys.argv[1:])

if args.check:
    check_list(args.file if args.file else sys.stdin.read().rstrip())
else:
    if args.file:
        with open(args.file,'rb') as f:
            print get_md5(f)
    else:
        print get_md5(sys.stdin)
...
```

briarfox:
@ywangd Added pipes but I'm unsure how to get around stdin.read() blocking if there is nothing in buffer. It waits for a keypress before exiting module.

briarfox:
Added sha1sum and sha256sum. I'm following the unix format but would it be better to just make a checksum to handle the various hashes or would the be best separate?

ywangd:
@briarfox If you program does not **by default** read from stdin, following Linux convention, a dash (`-`) is often used to represent stdin. 

For an example, `diff` does not **by default** read from stdin, therefore this command does not work `wget http://www.example.com/remote.txt | diff local.txt`. It will complain missing operand. So a dash has to be used to indicate the second file to `diff` is from stdin, like this `wget http://www.example.com/remote.txt | diff local.txt -`. 

I think this is also the situation with `md5sum`. You don't want it read from stdin by default, i.e. when it is called without any arguments, it exits immediately (with some help messages from argparse). So the code changes are:
* The `file` positional argument is reverted back to be compulsory
* Check whether the filename is `-`. If yes, read from stdin.
```Python
ap.add_argument('file',action='store',help='File to get md5sum or file contailing a list of hashes and files')
args = ap.parse_args(sys.argv[1:])

if args.check:
    check_list(args.file if args.file != '-' else sys.stdin.read().rstrip())
else:
    if args.file != '-':
        with open(args.file,'rb') as f:
            print get_md5(f)
    else:
        print get_md5(sys.stdin)
```

As for sha1sum and sha256sum, I personally always try abide to what Linux/Unix offers.It often seems to be easier to merge small utilities (by writing wrappers etc.) than splitting a large utility. But it is not really compulsory. If you feel it is more convenient to have a single checksum command, I am all good with it.

A few other comments (warning: wall of texts coming ~~) :
* Linux md5sum does seem to by default read from stdin.

* Current `md5sum` only takes a single filename when **NOT** reading from stdin, while it tries to handle multiple filenames when reading from stdin (using `str.split('  ')`). I personally think it shall handle arguments consistently no matter where they comes from. 

* According to Linux/Unix convention, stuff through pipes are most likely the contents of files, **NOT** the filename. e.g `cat afile | grep interesting`. The content of `afile` is sent through the pipe not its filename. If you want to send a filename through the pipe, it should be `echo afile | grep interesting`. Note this command is useless although `grep` does see the filename correctly. To really use this filename and make `grep` work with its content, the right Linux utility to use is `xargs`, which I also implemented for StaSh.

* To use xargs for the above example, you simply prepend the `grep` command with `xargs`, i.e.: `echo afile | xargs grep interesting`. The basic thing that `xargs` does is to call the following command (`grep` in this case) with all arguments plus whatever sent through the pipes. So in this case, `xargs` calls `grep interesting afile`, which is functionally the same as `cat afile | grep interesting`. 

* Use `xargs` is preferred and better than handle arguments yourself with `str.split(' ')`. For an example, following command will not work with current `md5sum`:
`echo 'space\ delimited\ filename.txt' | md5sum`
`md5sum` will treat it as three files while it really is just one single file. If `md5sum` wants to handle this case, you'll have to check escapes and quotes and soon you'll find yourself implementing another shell command parser. All these trouble can be saved with `xargs`:
`echo 'space\ delimited\ filename.txt' | xargs md5sum`
More powerful operations can be achieved such as:
```Bash
# Find all text files recursively and calculate md5sum for each of them
find . --name "*.txt" | xargs md5sum 
```
 No matter what comes through the pipe, `xargs` append them unchanged to the command to be called. Once the command is called, it is then again up to the shell to interpret the texts sent through the pipes, thus avoid re-inventing the wheel (i.e. shell parser). 

* So in summary, if texts from pipes are flienames that the command after pipe needs to open, `xargs` shall be used to construct the actual call. A command shall not worry about argument splitting, expanding or substitution, which are exactly what the shell does.

Hope I am making sense here. Sorry for the long texts.

briarfox:
@ywangd That was a great explanation. Thank you. I reworked all the files to allow for a straight stdin pipe and xargs. It now works on files and the stdin string. 

    cat md5sum.py | md5sum
    #equals
    md5sum md5sum.py

    #This works
    find . --name '*.py' | xargs md5sum
    
    #Quick way to compare a hash
    echo "f5d4ff9cf90dd3497542331029fe893d    pip.py" | md5sum -c

    #Check multipal files
    echo 'md5sum.py sha1sum.py sha256sum.py' | xargs md5sum

ywangd:
@briarfox Thanks again for the pull request! :)

PR #70, opened 2015-01-19
mail module

briarfox:
Simple mail module to send messages and files through stash. It's not ready for a pull accept but I'd like feedback.

### DIFF
--- a/bin/mail.py
+++ b/bin/mail.py
@@ -1,0 +1,177 @@
+'''
+mail - send messages/files from stash.
+Calling mail without params will prompt for input.
+Passing '-' in the message param will read from stdin pipe.
+
+usage: mail.py [-h] [-s SUBJECT] [-v] [-f FILE] [-e] [sendto] [message]
+
+positional arguments:
+  sendto                Send email to.
+  message               Email Message. Passing '-' will pass stdin from pipe.
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -s SUBJECT, --subject SUBJECT
+                        Email Subject.
+  -v, --verbose         Verbose print
+  -f FILE, --file FILE  Attachment to send.
+  -e                    Edit .mailrc
+'''
+
+import os
+import smtplib
+import argparse
+import sys
+from ConfigParser import RawConfigParser
+ 
+from email import Encoders
+from email.MIMEBase import MIMEBase
+from email.MIMEMultipart import MIMEMultipart
+from email.Utils import formatdate
+from email.MIMEText import MIMEText
+ 
+class Mail(object):
+    def __init__(self,cfg_file='',verbose=False):
+        #from config
+        self.cfg_file = cfg_file
+        self.verbose  = verbose
+        self.user     = ''
+        self.passwd   = ''
+        self.auth     = False
+        self.mailfrom = ''
+        self.host     = 'smtp.fakehost.com'
+        self.port     = 537
+        self.tls      = False
+        self.read_cfg()
+        
+    def _print(self,msg):
+        if self.verbose:
+            print msg
+            
+    def read_cfg(self):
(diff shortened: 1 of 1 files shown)


ywangd:
@briarfox Thanks for the pull request.

I had couple tries but didn't get success with my gmail account. I am not sure what is the recommended settings for `.mailrc` for a gmail account. The closest one I got is that Google blocked the signin access and emailed me that I have to lower my account security settings before the signin can work.

Another minor issue is that default cfg reading has interpolation on so that directives like `%(foo)` can be interpolated. I think this feature should be off and read raw values only.

briarfox:
@ywangd It sounds like you are using two step verification. You'll need to go to security settings and generate an app passwor]d (used for apps that don't use two step). Let me know if you get it working. I'll have a look at the cfg again.

briarfox:
I seem to have greated a merge confict with .gitignore. I believe I accidently rebased off master instead of dev. I'm unsure how to resolve merge conflicts. Should I create a fresh pull request?

ywangd:
@briarfox Did you already re-send the pull request. It seems clean to me now. I modified .gitignore yesterday to ignore the record file for scsm (stash-command-script-manager).  This may be the cause of the conflict. It should haven been easy to handle on my side. You can add back the `.mailrc` ignore now.

As for the EOF error, it was in fact a bug in stash's readline method. It is now fixed. Thanks a lot for finding the bug.

briarfox:
@ywangd Is selfupdate suppose to update stash.py or just the /bin? I've manually pasted in the dev stash.py in the past. Just tried selfupdate (using dev) and still getting EOF error.

I'll make the changes to prompt input and submit pull in a few hours.

ywangd:
It updates stash.py as well. You need restart stash for the changes to
become effective.

Updates to command scripts do not require restart.

On Thursday, 22 January 2015, Chouser <notifications@github.com> wrote:

> @ywangd <https://github.com/ywangd> Is selfupdate suppose to update
> stash.py or just the /bin? I've manually pasted in the dev stash.py in the
> past. Just tried selfupdate (using dev) and still getting EOF error.
>
> I'll make the changes to prompt input and submit pull in a few hours.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ywangd/stash/pull/70#issuecomment-70850462>.
>


-- 
----------------------------------------------------------------------
                      Life is elsewhere

briarfox:
@ywangd The fix to raw_input seems to have broken part of ssh.py. I'm not getting \n added to input text. Example: 

    connect to ssh server. 
    pythonista put mail.py mail.p

This transfers "mail.py\n" an ls shows it as "mail.py?"

I'll check and make sure it isn't an issue with ssh.

briarfox:
@ywangd I think i figured it out. I believe before your fix stash was scrubbing \n (LF) from stdin. Is this correct? The ssh module reads directly from stdin. Now it's getting a LF added in. Should be an easy fix.

ywangd:
@briarfox I made the fix to raw_input to be consistent with default Python file-like object behavior. But it breaks something that `ssh` relies on. 

Python `raw_input()` automatically strips the ending `\n` if there is one. But `readline()` method returns the `\n` character unchanged. Previously I made `readline()` also strip the ending `\n`, which `ssh` relies on but this behavior is not consistent with other Python file-like objects.

So I'd like to suggest a minor change to `ssh` to either strip the ending `\n' or use `raw_input()` instead of `sys.stdin.readline()`

briarfox:
@ywangd I think mail is good to go unless you have any suggestions. Were you ever able to get gmail working? Also submited an ssh fix #72

ywangd:
@briarfox I did get gmail working by change settings to "allow lesser secure app". Thanks!

PR #71, opened 2015-01-20
Fixed missing imports in git

jsbain:
This fixes missing imports.  As a warning, in some cases you will get a broken pipe error.  This seems to be caused by commits being made locally and on the server, such that fast forward is not possible.  The solution is to clone the repo again, make changes, then push.  I'm working on merge capability for gitview, which will be portable to stash and hopefully will take cRe of the issue

### DIFF
--- a/bin/git.py
+++ b/bin/git.py
@@ -24,6 +24,10 @@ commands:
 
 SAVE_PASSWORDS = True
 
+import argparse
+import getpass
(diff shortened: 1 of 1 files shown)


ywangd:
@jsbain Thanks a lot for the fix. Being able to do recursive merge will be nice. Personally I am pretty satisfied already with fast-forward merge. I'll give it some more tests.

PR #74, opened 2015-01-23
Cursor positioning with swipe gesture

jsbain:
Here is a cursor positioning via swipe gesture on the virtual keyboard, so it acts like the pythonista VK.  Basically, the VK now holds a ScrollView, which allows capturing of swipes without switching back to the editor( unlike a custom ui.View).  Except, rather than scrolling the ScrollView, it keeps track of incremental changes, applying a scale factor from pixels to characters.  When the integrated offset is more than 1 character, the ShTerm.io cursor position is updated.

In retrospect, probably only one ShVk is needed, but I created one for each key group, since originally things were not working.  also, I suspect that you will want to add higher level cursor positioning functions to ShTerm, rather than the low level TextField manipulation that I resort to here.  Not sure if this would work with ssh for example.  But this works nicely, and I think is easier than an actual arrow button.  Vertical scroll would also work (would need to oversize the `content_size` height, and track vertical motion, perhaps calling the VK tapped functions for up and down buttons to scroll through history..


ywangd:
wow thanks @jsbain !
I only had the new branch pushed couple hours back. Here you already sent a pull request and it looks to be a really useful feature. 

I am still looking at your changes and learning how it is done. I do notice that the swipe is only recognised when swiping on an area that is not covered by those virtual buttons. This makes it difficult to use on my iPhone 4s unless I put on the landscape mode. So it would be great if swipe can also be tracked when on top of those buttons (Pythonista's own VK does have this function). Is it possible?

I totally agree that higher level interfaces are needed for cursor position manipulation. There are many things needs to be done to rationalise interfaces, especially around ShTerm. I'll keep working on them.

PS: I am going to manually merge this request as Travis CI test failed. But that is only because the dummyui module does not yet have a ScrollView class. It will be an easy fix. Thanks!

jsbain:
If you start swiping in a non button area, it should recognize the swipe.  Actually, on my ipad, I thought that even if I started on a button, if I immediately swipe, it recognized it as a swipe.  I'll play around with it later.

ywangd:
Sorry jsbain. You are right. It does recognise the swipe even on the button
area!! That's awesome! Thanks!

On Saturday, 24 January 2015, jsbain <notifications@github.com> wrote:

> If you start swiping in a non button area, it should recognize the swipe.
> Actually, on my ipad, I thought that even if I started on a button, if I
> immediately swipe, it recognized it as a swipe. I'll play around with it
> later.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ywangd/stash/pull/74#issuecomment-71298818>.
>


-- 
----------------------------------------------------------------------
                      Life is elsewhere

ywangd:
@jsbain The popup button looks awesome. Is it possible to have popup buttons and still retain the swipe gesture for cursor positioning? Or will the popup button intercept the swipe gesture and block the scrollview from working?

It would be great if we can have both popup buttons and swipe gesture support. But even if they cannot co-exist, it will still be useful to have at least one popup button in stash. I am thinking of the rightmost one that by default switch key groups. A long-tap can bring up a mini-menu and many interesting stuff can be put in there, e.g. configuration management, possible future multi-tab switch, buttons for possible future editor mode.

jsbain:
We're somewhat limited in what we can do for ui's presented as `panel`, because it means i don't get to use touch_moved.

With minor changes to the PopupButton concept, it should be possible to implement a simply long-tap popup group of buttons.
I'll have to verify that the scroll gesture doesn't interfere with the touch events for custom buttons.
The pop up will work differently than my example, because in that example you drag your finger to select the button:  this will actually be much simpler, long tap will pop up a list of other characters in an extra row (or maybe could fly out as a vertical column, or depending on orientation), which you then would use a second tap to either select a key, or tap outside of the VK to cancel entry.  I'll mock something up this weekend.

ywangd:
@jsbain Yes It is indeed annoying that drag gesture cannot be used in *panel* mode. I hope this is a bug and will be fixed eventually. In my opinion, *panel* mode is still the best choice for stash even with its limitation simply because it persists. 

You are right that what I described really is just a long tap action. I didn't realise the difference. I guess *drag and select* will not work in *panel* UI. So long tap is an Ok alternative. I'd prefer to keep the cursor positioning function. So if custom view does interfere with it, we can just have one custom view button at the rightmost position.

PR #76, opened 2015-01-24
initial crypt.py

briarfox:
crypt is a little file encryption module that uses AES in CBC mode to encrypt files. I'm contemplating adding public keys to encrypt the random passphrase for AES and including it with the encrypted file. I'm hoping to add more features but the AES portion seems to be working well.


ywangd:
@briarfox Thanks again for pull requests! I am about to tag v0.3.0 and move it to the master branch. So I'll save this request for v0.4.0.

PR #82, opened 2015-01-27
using simplecrypto pypi package

briarfox:
Moved to using simplecrypto package (wrapper for pycrypto). Encryption key method has changed so any previous encrypted files will be broke. I'm still working on adding pub key encryption but you can accept this pull request if you like.


ywangd:
Thanks for the update!

PR #87, opened 2015-01-30
Very simple telnet client.

jsbain:
this is not a good long term solution, but someone requested telnet, and I wrote one some time ago.


ywangd:
Thanks @jsbain 
This will be a good start point.

PR #88, opened 2015-01-30
added git fetch command.   updated handling of remote shortcuts

jsbain:
fixed handling of `git push remote` style entries, using remote shortcut
rather than full uri.
git remote command can now add new remote shortcut


ywangd:
Thanks for the continuous support on this. It seems to solve the puzzle polymerchm was having.

PR #97, opened 2015-02-05
simple telnet

briarfox:
I've been playing with twisted and wanted an easy way to interface with the server. Here is a simple telnet client.

### DIFF
--- a/bin/telnet.py
+++ b/bin/telnet.py
@@ -1,63 +1,123 @@
-''' telnet hostname [port] [timeout]
-
-a very simple telnet client.
-
-
 '''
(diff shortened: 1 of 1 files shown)


ywangd:
@briarfox I assume this is an update to the `telnet` script that @jsbain put in a few days back. It looks to me that the screen clear is removed. Is that something necessary for watching ASCII Star Wars?

briarfox:
@ywangd And thats what I get for working on a module before pulling new changes. It's a stand alone module. I didn't see the other telnet one. I'll remove pull request.

PR #99, opened 2015-02-06
fixed <CR> and removed prints

briarfox:
Removed testing prints

Now sends carriage return


ywangd:
@briarfox Thanks for working on this

PR #105, opened 2015-02-14
man without arguments: show summaries 

oefe:
So far, `man` without argument just listed the available commands. This patch adds a brief summary of the command description after each command. This should be helpful in particular for movie users, and users unfamiliar with posix commands:

```
[~/Documents]$ man
     alias: List or define shell aliases.
       cat: Print the contents of the given files.
        cd: Change the current working directory.
...
```

## Notes: ##

- The first line of the docstring is used as summary. This is in accordance to  [PEP 257](https://www.python.org/dev/peps/pep-0257/)
- I updated several docstrings to provide better summaries.
- This parses all Python scripts in BINPATH. With many scripts, this could in theory be slow. However, with the commands currently included with stash, it's still quite fast (on my iPad 4). If necessary, performance could be improved by caching the docstrings.


ywangd:
Thanks @oefe 
These are good changes and well explained.

Issue #106, opened 2015-02-15
git: switch to jsbain fork of dulwich 


oefe:
In the `dev` branch, `git fetch` fails with `AttributeError: 'module' object has no attribute 'fetch'`

The reason is that it is still using the `transistor1` fork, which hasn't been updated in a while, and doesn't contain `fetch`.

oefe:
Guess this should be closed...

PR #107, opened 2015-02-15
Switch to the jsbain fork of dulwich (issue #106)

oefe:
This is an attempt at fixing issue #106 

Unfortunately, it is not as simple as exchanging the URL. We need also to check wether we are using the old version (by checking whether `fetch` is available), clean up the old versions, and reload `dulwich`

### DIFF
--- a/bin/git.py
+++ b/bin/git.py
@@ -319,6 +319,20 @@ def git_reset(args):
             with open(str(path),'w') as f:
                 f.write(file_contents)
 
+def get_config_or_prompt(repo, section, name, prompt, save=None):
+    config = repo.repo.get_config()
+    try:
+        value = config.get(section, name)
+    except KeyError:
+        value = raw_input(prompt)
+        if save == None:
+            reply = raw_input('Save this setting? [y/n]')
+            save = reply == 'y'
+        if save:
+            config.set(section, name, value)
+            config.write_to_path()
+    return value
+        
 def git_commit(args):
     ap=argparse.ArgumentParser('Commit current working tree.')
     ap.add_argument('message',default=None,nargs='?')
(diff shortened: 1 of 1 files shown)


oefe:
Oops, I committed another change in this branch that should probably go into a separate pull request. 

Let me know if I should separate these changes

oefe:
Another (completely unrelated) thought: would it be better to make pull requests for the git command against @jsbain 's repository?

ywangd:
@jsbain Could you please have a look at these changes since you are the author the `git` command?
@oefe I agree it is better to make pull requests for `git` command to @jsbain 's fork. Does it sound reasonable to you, @jsbain ?

jsbain:
Either way re pull.   I'd suggest checking `dulwich.__version__`, rather than existence of `fetch`.  Transistor1 version was 0,9,8.  I updated my fork to be `(0,9,9,'jsbain_fork')`.   My eventual goal is to update the code in the git module to use the official dulwich versions, since there is really nothing needed in the fork ... Need to just update a few method signatures to use the original.  

Checking the version tuple will allow an easier way to specify which form we need.

jsbain:
Ok, I've merged this change, along with a change to check for version.  This ended up being more difficult than I thought, because it is possible someone has dulwich in another folder, so it is necessary to delete all dulwich packages from sys.modules.  In retrospect, we should be deleting the libpath from sys.path after using it, and maybe importing dulwich using an alias, so that it doesn't stomp on anyone else's version.

I'll check in as soon as I detangle the mess I created when using merge in the strange state that resulted...

jsbain:
https://github.com/ywangd/stash/pull/109

PR #109, opened 2015-02-16
Incorporate oefe's changes, and some other minor fixes

jsbain:
Incorporate oefe'self fixes (issue #106 ), using dulwich version as the check.  This also fixes some problems that cropped up with this approach, if user already had dulwich installed elsewhere.

This pull also adds git reset --soft (change head pointer only).  This is the default right now, but that may change, I need to verify how real git works.  also, I added a method to reload the current file in the editor when doing a checkout, since previously the old file got stuck and caused confusion.

Finally, fixed an issue with git merge where it didn't update the MERGE_HEAD, etc if the merge was successful.

### DIFF
--- a/bin/crypt.py
+++ b/bin/crypt.py
@@ -1,5 +1,5 @@
 '''
-crypt - File encryption for stash
+File encryption for stash
 Uses AES in CBC mode.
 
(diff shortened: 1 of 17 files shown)


ywangd:
Thanks @jsbain and @oefe

PR #116, opened 2015-04-06
better error handling if current directory isn't a repository

oefe:
If the current directory is not a repository, most git commands fail with a `StopIteration` error. This PR tries to fix that and issue a better error message


jsbain:
Looks good to me.

PR #123, opened 2015-08-15
Implementation of Wake on LAN

GeorgViehoever:
wol (wake on lan) command for stash, based on
https://github.com/remcohaszing/pywakeonlan.
Tested in my local network with Pythonista 1.6 beta,
but it should also work with 1.5

### DIFF
--- a/bin/wol.py
+++ b/bin/wol.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+# Based on https://github.com/remcohaszing/pywakeonlan as of 2015-08-13
+# Adapted and extended by Georg Viehoever, 2015-08-13
+#
+# License as in original:
+#            DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
+#                    Version 2, December 2004
+#
+# Copyright (C) 2012 Remco Haszing <remcohaszing@gmail.com>
+#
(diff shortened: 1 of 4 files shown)


bin/wol.py
```diff
@@ -0,0 +1,76 @@
...
+import os.path
+
+# manipulate sys.path to include stash lib directory (where wakeonlan is found)
+# and have stdlib before .
+pythonLibPath=os.path.join(os.path.split(sys.executable)[0],'lib')
+sys.path.insert(0,pythonLibPath)
+
+try:
+    # while running in stash environment
+    stashLibPath=os.path.join(_stash.runtime.envars['STASH_ROOT'] ,'lib')
```
ywangd: `os.envrion['STASH_ROOT']` should work here as well.

ywangd:
Thanks for the pull request. Especially thanks for updating the README file.

StaSh recognises a `PYTHONPATH` environment variable and adds its contents to `sys.path`. Currently, it is NOT  defined by default. But now I think it is helpful to set the default value to `$STASH_ROOT/lib` so that running scripts (e.g. `wol.py`) do not have to manage the import path.

Once the `PYTHONPATH` change is made, I'll modify `wol.py` accordingly. I hope this is OK with you.

GeorgViehoever:
Hi,
your changes are perfectly fine for me. Thanks! I am new to Pythonista and stash, so help from the experienced developers is more than welcome.
Georg

PR #124, opened 2015-08-16
improved handling of sys.path

GeorgViehoever:
Better manipulation of sys.path, in particular avoiding constant growth

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,17 +160,11 @@ The usage of StaSh is in principle similar to Bash. A few things to note are:
   `~/Documents/bin:$STASH_ROOT/bin`.
 
 * The executable files are either Python scripts or StaSh scripts. The type of
-  script is determined by first looking at the first line of the file - following the common
-  UNIX/Linux convention of determining the type of a script by specifying a line
-  with `#!pathToInterpreterExecutable`. If the first line contains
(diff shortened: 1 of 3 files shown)


GeorgViehoever:
Also added ability to analyze #! lines to determine script type

ywangd:
I have made some changes so `sys.path` is intact after script run. The `lib` sub-folder is now part of the default `sys.path` so it is no longer necessary to handle it in `wol.py`, i.e. you can directly run `from wakeonlan import wol` and it will work.

As for checking the shebang line, there were some discussions before. I personally think it is an overkill to perform the check. Pythonista is a very controlled environment. Unless you make extra effort, all python scripts are by default suffixed by `.py`. And `.sh` files are rare given the very limited scripting ability StaSh provides. So I'd like to assume who creates `.sh` files really understand what he/she is doing. All in all, it is a pretty good approximation to tell the script type by the file name suffix. Also I am unsure about the performance if StaSh needs to open and read every file to tell its type. This is especially a concern for the Tab Completion suggestions. It again relies on filename suffix for file types so it can search through a directory much faster compared to open/read files.

GeorgViehoever:
removed checking for #! in stash.py, adapted to your new sys.path handling. Minor changes in README.md

ywangd:
Thanks for making the changes!

PR #139, opened 2016-01-05
Git fixes

jsbain:
This fixes a variety of issues with git in the beta.  I have converted over to production dulwich, pulling upstream changes to include the features we needed.  (eventually this will just point to jelmer/dulwich, but for now, i am still pointing to my fork).   

Also, I included here (sorry, should have been a separate pull) a change to launch stash which keeps the stash variable from getting cleared.  Basically, naming a module with a double underscore keeps it safe across global clears (see the pythonista preflight script).

Also, in keyboard_frame_did_change, only take action when the view is on screen.  This prevents problems when trying to kill stash, but it keeps running and generates errors whenever the kb frame changes.  Also, I think this fixes the issue on startup sometimes where the frame gets all hosed, and ends up being too small.


ywangd:
Thanks for the pull request.

With the change to `launch_stash.py`, the module reload part now reads as follows:

```Python
if '__stash' in sys.modules:
    stash = sys.modules['__stash']
    reload(__stash)  # reload to ensure any changes to be honoured
```

I wonder whether the variable `stash` in the second line should be change to `__stash` as well so that it reads as `__stash = sys.modules['__stash']`

jsbain:
My thought was to minimize any disruptions to other code that used import stash, or expected a  module named stash.... this way there is still a stash in globals.  Either way is ok by me, I was continuing to have trouble with global clears when stash is not in site-packages, and this solves it.

ywangd:
The `beta16` branch has stash installed as part of `site-packages`. I guess it will solve this issue (global clearing) as well?
Hopefully the new Pythonista will come out soon so that stash can also be updated to the new version (which drops 1.5 support).

jsbain:
I had mixed results with beta16 (was that the highlighted version?) though maybe i should give it another go.

ywangd:
beta16 is the version with colors (https://github.com/ywangd/stash/tree/beta16). Did you experience much crash? Some details will be very helpful.

The ObjC module enables many great things that's not possible in 1.5. So beta16 will become the new master once Pythonista 2.0 hits the appstore. Any help to make it more stable is highly appreciated.

PR #145, opened 2016-01-23
Open file in new editor tab by default

Dutcho:
With Pythonista 2.0 out, it's now possible to specify a file must be opened in a separate editor tab. This is desirable in the common use case, so I changed to make this the default behaviour, with an '--old_tab' option to reuse an existing editor tab (if any).
Also, while the argument specification made the path argument optional, this wasn't processed correctly. It should work correctly in all cases now (including some special casing if a path is specified neither at the start nor when saving a temporary file).


ywangd:
Open in new tab by default is a great improvement for the Pythonista 2.0. Thanks for the pull request!

PR #146, opened 2016-01-27
Simplify things when find_script_file() fails

cclauss:


ywangd:
As discussed in #147, the `fullname` variable is in fact necessary and cannot be replaced by the modified code. Thanks!

cclauss:
Now I understand.  Thanks for the explanation.

PR #173, opened 2016-04-17
python command: added interactive console

bennr01:
Changes to the python command:
   -added a interactive console (started when not called with "-c","-m" or file argument)
   -changed the way the command-line arguments are handled.
   -changed the help-paramters to be more like CPython
   -fixed a bug allowing to pass both "-c" and "-d" (it now prints an error-message)
  -converted spaces to tabs (i edited this file in pythonista, it is annoying to see a mixture of tabs and spaces)


ywangd:
Thanks for the pull request.

PR #176, opened 2016-04-25
added new commands

bennr01:
!This pull-request may contain english spoken by a non-native english-speaker!

I have added two new commands: gci and mc.

mc:
-name reffers to the command-line-tool "midnight commander" (this is definitly not a full counterpart)
-allows browsing files on other filesystems (currently supported: local, FTP and dropbox)
-allows browsing files on multiple filesystems synchronously.
-designed to be easily extendable (simply subclass BaseFSI and add it to INTERFACES)
-allows reading files on other filesystems
-allows copying and moving files between two filesystems (even between dropbox and FTP)
-allows "running" a command on a different filesystem (download files into a tempdir, cd into it, run command, cd back, upload files if changed)
-uses an internal command-loop (using the cmd-module)
-contains help
-contains a dropbox setup assistant
-allows running a stash-command normally while beeing in the command loop (using _stash(cmd))
-implements common commands directly (e.g. cat and mkdir)
-still has a few minor bugs left which i dont know how to fix (e.g. dropbox-interface freezes sometimes, this may be caused by a ddos-protection)
=>I wrote a bug and todo list at the beginning of the file, you may want to read it before merging (maybe you think it is to buggy)

gci:
-interface to the garbage collector
-subcommands allows enabling/disabling the garbage collector
-subcommands allows getting/setting debug and threshold
-subcommand for viewing stats (RAM usage of objects, number of tracked objects and garbage, RAM usage of garbage)
-subcommand for fixing memory leaks (this only works sometimes, reference cycle breaking is hard)
-usefull for debugging
-usefull for tweaking tests

Both files are not PEP8-compiliant as the code would be even harder to read on a small device (at least i think so).

And a short question: would it be possible to add a command starting a subshell and exiting (stash(cmd) returns) when the "exit"-command is used? This may be helpful in combination with the
mc.py script.

### DIFF
--- a/bin/gci.py
+++ b/bin/gci.py
@@ -1,0 +1,59 @@
+# coding: utf-8
+"""interface to pythons built-in garbage collector
+
+Warning: this command may crash StaSh. Only use it if you know what dou are doing!"""
+import gc,argparse,sys
+
+def main():
+	parser=argparse.ArgumentParser(description="access to pythons built-in garbage collector")
+	parser.add_argument("command",help="what to do",choices=["enable","disable","status","collect","threshold","debug"],action="store")
+	parser.add_argument("args",help="argument for command",action="store",nargs="*")
+	ns=parser.parse_args()
+	if ns.command=="enable":
+		gc.enable()
+	elif ns.command=="disable":
+		gc.disable()
+	elif ns.command=="collect":
+		gc.collect()
+	elif ns.command=="status":
+		print "GC enabled:              {s}".format(s=gc.isenabled())
+		tracked=gc.get_objects()
+		n=len(tracked)
+		print "Tracked objects:         {n}".format(n=n)
+		size=sum([sys.getsizeof(e) for e in tracked])
+		del tracked#this list may be big, better delete it
+		print "Size of tracked objects: {s} bytes".format(s=size)
+		print "Garbage:                 {n}".format(n=len(gc.garbage))
+		gsize=sum([sys.getsizeof(e) for e in gc.garbage])
+		print "Size of garbage:         {s} bytes".format(s=gsize)
+		print "Debug:                   {d}".format(d=gc.get_debug())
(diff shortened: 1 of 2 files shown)


ywangd:
Thanks for the pull request. The `mc.py` command is almost the size of the original `stash.py`!

Technically, each command is running on its own sub-shell. When you use `_stash(...)` in the script, it starts another further down level sub-shell. I am not quite sure if I understand your question. What exactly do you want achieve?

PS: I haven't had time to work on StaSh recently. Sorry for the delayed response.

bennr01:
The mc.py command allows running a stash-command "on" a remote filesystem by first downloading the specified files into a temdpir, then tell stash to cd into the tempdir and runs the command using _stash(). After _stash() returns, it tells stash to cd back to the last path and upload the files. Currently, this this subcommand allows running a single line (e.g. "unzip") before reuploading the files. However, if it would be possible to launch a subshell with human input-loop (like the normal stash shell where the user inputs the commands / the normal command loop when no scrpt is running) which would block _stash() until the user typed "exit", this may be a good alternative to a mount command.
I mean someting like the "cmd" command in window's cmd or like the "sh"-command in a linux-shell.
This would let the mc command do something like this:
-download all dropbox files into a tempdir
-cd into tempdir
-start a subshell using _stash()
-the user sees a normal stash commandline (where he can run commands like python, git, ...). _stash would block until the user uses the "exit" command. until he does this, he could work normaly with stash in this dir (unzip, tar, wc, redirect output, anything you can do with stash)
-_stash() returns once the user types the "exit" command
-mc reupload the files into the tempdir
With other words: a pseudo mount command
In the pull-request, i said subshell but meant something like a new instance of the stash user-interface (still the same I/O, but the script would be blocked by calling the command ant the user would regain controll of the stash console until he types "exit").

PS: Sorry for my horrible english skills.

PR #181, opened 2016-06-16
use https, as now required by pypi

jsbain:
pypi now requires https.


ywangd:
@jsbain It seems that you made this change to an old version of pip and is very difficult to merge. Could you please make a new pull request using the latest version and to dev? Thanks!

PR #182, opened 2016-06-16
fix https in pip

jsbain:


ywangd:
hmm... there is no change in this pull request??

jsbain:
that was strange.  anyway, fixed it now

PR #183, opened 2016-06-16
New command, bugfixes and improvements

bennr01:
This pull request contains mostly minor bugfixes and improvements.

**New command**: `webviewer`
   -opens the webbrowser without freezing StaSh
   -implemented to fix `mc`
   -named webviewer instead of webbrowser to prevent naming conflicts with the module
**Improvements**:
   -`mc.py` and `gci.py` are now compatible with PEP8
   -`mc.py` and `gci.py` now uses colors
   -more autocompletion for commands
   -`mc.py`: now supports non-integer IDs
   -`mc.py`: now supports dropbox logout
   -(!) `selfupdate.py` and `getstash.py`: now support updating from other users forks
   -added a new tip to `.stash_tips.py`
**Bugfixes**:
   -fixed some bugs in `mc.py` (dropbox setup should no longer freeze StaSh. Sorry, missed that bug before as i tested the dropbox setup outside StaSh)

**WARNING**: `selfupdate.py` and `getstash.py` has been modified. I tested it on different devices and with both Pythonista 2 (but not 2.1) and 3 and it seems to work without bugs. For more info about the tests, see comment on https://github.com/bennr01/stash/commit/fc4862baebce49766a9fcb41328aa0312f82dc5a . The comment will probably be auto-appended to this commit anyway. I could not test the one-line installation, but the link seems to redirect to the getstash.py file on `ywangd/master`, so it should work.

### DIFF
--- a/.stash_tips
+++ b/.stash_tips
@@ -28,5 +28,6 @@
     "Press the Tab button to auto-complete command and file names",
     "Show detailed traceback when a script fails with stashconf py_traceback 1",
     "Show a random tip with command totd",
-    "Edit files with unconventional names using edit, e.g. edit .stash_config"
+    "Edit files with unconventional names using edit, e.g. edit .stash_config",
+    "Use mc to access files in your Dropbox or on a FTP-Server"
 ]
--- a/bin/gci.py
+++ b/bin/gci.py
@@ -1,75 +1,104 @@
 # coding: utf-8
 """Interface to pythons built-in garbage collector
 
-Warning: this command may crash StaSh. Only use it if you know what you are doing!"""
-import gc,argparse,sys
+Warning: this command may crash StaSh!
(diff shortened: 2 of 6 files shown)


bennr01:
@ywangd Using `repo/branch` is probably better than how i made it. I made it this way to minimalize the changes to `selfupdate.py` and `getstash.py` and also to make it easier for new users . Maybe also auto-replace ":" with "/" in the argument (`bennr01:dev` would become `bennr01/dev`)? This would allow the use of a GitHub-like style for the repo/branch argument.

ywangd:
I have made the changes. The command now takes either `owner/branch` or `owner:branch` as the update target. It is currently in the `dev` branch and I'll merge it to `master` in a day or two. Thanks!

PR #187, opened 2016-07-05
Fix ShIO.readlines()

glider-gun:
This fixes ShIO.readlines() which returned entire buffer string, instead of list of lines.


ywangd:
Thanks!

Issue #179, opened 2016-06-14
Pythonista3


syslot:
Pythonista was released，would you prepare to adapt to it with python3?

ywangd:
You can still use StaSh with the new release. The `launch_stash.py` script contains a `python2` shebang that will allow it running in Python 2 mode even when you have Python 3 as default interpreter.
 
With above being said, I do plan to make stash python 3 compatible. In fact, the [`py3`](https://github.com/ywangd/stash/tree/py3) branch exists for this very purpose. However the work is not completed as there are some issues porting the various command scripts, especially `git` as one of its dependency is Python 2 only. 

Since new Pythonista can run both 2 & 3 code, I am thinking to port only the StaSh main programs to Python 3 and leaves the command scripts to Python 2. However, one critical missing piece is to invoke the corresponding Python interpreter from with StaSh based on the shebang line of a script. This is going to involve some objc hacks and I haven't had time to work on it. @jsbain could you please shed some light on this?

jsbain:
I think i was able to execute a script that was already loaded in the editor, but not a generic script.   I forget.  I think I concluded it was going to be difficult to do it seamlessly from within a script.  But there may also be a way from a cdll, I have not played with that, not sure if both are accessible... 
>

syslot:
Well ,  very thanks to your tool, and stash is real nice project!

ywangd:
@jsbain 
I tried with `PA3PythonInterpreter` object and it can run an arbitrary script using the `runScriptWithFileAtPath_argv_` method. I am not sure how it would work when putting in a separate thread though and that's something to test. 
However, I was NOT able to get the `PythonInterpreter` object to work. There is this `runFile_argv_captureExceptions_completion_` method, but I don't know what do you pass as the third and fourth arguments. 

Since I plan to make StaSh main programs PY3 compatible, it is more important to get `PythonInterpreter` to work as most of the command scripts are PY2.

syslot:
I try ssh & gh in pythonista3 and two scripts are broken.

jsbain:
I Have gotten this to work through the editor tab:
https://gist.github.com/e64581caa4588625896594882870c50c
The file needn't be open, but some file needs to be open (tab must exist).  If interpreter is already running, you must wait for running to end or your request is ignored.  See code to do so (not thread safe, but a Lock could work).  
The main drawback-- globals are cleared.

[...]

Issue #118, opened 2015-05-07
Adding Vim


briarfox:
Just found a pure python implementation of vim, It looks like it would be a great addition to stash, i'll see if I can get it working on the ipad. If anyone is interested: https://github.com/jonathanslenders/pyvim

ywangd:
This is an interesting project. But I suspect it would in Pythonista as it most likely requires modules like `termios` for manipulating screen text. I could be wrong. Please let me know your findings. Thanks!

dmkanter:
+1 love the idea. I actually attempted to install it using pip, but have no idea where it ended up in the filesystem.

ywangd:
`pip` installs all packages to the `site-packages` folder.

dmkanter:
Ok, I see it there, just not the actual executable I'm supposed to run to launch it.

Seanld:
`vim` is a text editor that has runs on a binary executable. You won't be able to run it in Pythonista unless you make a port of it to Python. The other issue is that I don't think StaSh has a cell-manipulation system where you can color cells on the shell like pixels on a screen, which is how Vim operates.

Seanld:
Since you found a pure Python version, it should work. The only thing I can think of where it wouldn't work is StaSh's stdout. I don't know if it will be able to handle Vim's cell UI manipulation.

PR #190, opened 2016-07-08
New command, bugfixes and improvements

bennr01:
Yet another bunch of bugfixes for the `mc`-command, but also some improvements to the python-command.

**New Command and Files**
- `ftpserver` uses pyftpdlib to start a ftpserver. I frequently use omz`s ftpserver and tought it may be usefull to add to StaSh.
- added a helppage for the manpages. I tought it may be useful to document the behavior of the new `man`-command (in case others want to write manpages; i dont think my english is sufficent to write anything wich is not to hard to read)

**Improvements**
- The `python` command will now pass `_stash` and `__debug__` to the interactive interpreter.
- cleaned up the code of the `mc`-command by removing all unnecessary assignments in `except-clauses`.

**Bugfixes**
- Fixed a bug in `mc` where menu-entries were swapped. I accidently added this when cleaning up the menu-entries in `mc`  in an earlier commit.
- Fixed a bug in `mc` where no help was shown for a topic.

**Short info**
- I have updated the `getstash.py`-file in my fork, so testing this changes with `selfupdate -f bennr01:dev` should work now. I recently noticed that `seldupdate -f user:branch` wont work if the fork still contains the old `getstash.py`.

Also, a question: What do you think about the idea of adding a command for enabling and disabling monkeypatches for modules and functions not working in Pythonista? For example, patching `subprocessing` and `os.popen` to call the shell-commands in StaSh. The functions and modules i thought about patching dont work in Pythonista, so this could make more "normal" python-scripts compatible with Pythonista by using StaSh. Also, because they dont work by default, any weird behavior caused by monkeypatching should be acceptable because the scripts would still work more than by not doing any monkeypatching. The monkey-patching would be done by a command, which would have to be run by the user everytime he wants to enable them (they would be disabled by default and resetted whn Pythonista restarts).

### DIFF
--- a/bin/ftpserver.py
+++ b/bin/ftpserver.py
@@ -1,0 +1,95 @@
+"""Simple FTP Server"""
+import argparse
+import os
+import sys
+import threading
+import time
+import logging
+
+_stash = globals()["_stash"]
+
+try:
+	import pyftpdlib
+except ImportError:
+	print("Installing pyftpdlib...")
+	_stash("pip install pyftpdlib")
+	es = os.getenv("?")
+	if es != 0:
+		print(_stash.text_color("Failed to install pyftpdlib!", "red"))
+		sys.exit(1)
+
+from pyftpdlib.authorizers import DummyAuthorizer
+from pyftpdlib.servers import FTPServer
+from pyftpdlib.handlers import FTPHandler
+
+
+def run(ns):
+	"""starts the server."""
+	auth = DummyAuthorizer()
+	if ns.user is not None:
+		auth.add_user(ns.user, ns.pswd, ns.path, perm=ns.perm)
+	else:
+		auth.add_anonymous(ns.path, perm=ns.perm)
+	handler = FTPHandler
+	handler.authorizer = auth
+	handler.banner = "StaSh v{v} FTP-Server".format(v=_stash.__version__)
+	address = ("0.0.0.0", ns.port)
+	server = FTPServer(address, handler)
+	server.max_cons = 128
+	server.max_cons_per_ip = 128
+	# setup logging
+	logger = logging.getLogger("pyftpdlib")
+	logger.setLevel(logging.CRITICAL)
+	logger.propagate = False
+	# server needs to run in a thread to be killable
+	thr = threading.Thread(
+		name="FTP-Server Thread", target=server.serve_forever
+		)
+	thr.daemon = True
+	thr.start()
+	print("FTP-Server started on {h}:{p}".format(h=address[0], p=str(address[1])))
+	try:
+		while True:
+			time.sleep(0.2)
+	except KeyboardInterrupt:
+		print("Stopping Server...")
+		server.close_all()
+
+if __name__ == "__main__":
+	parser = argparse.ArgumentParser(description=__doc__)
+	parser.add_argument(
+		"-p", "--port", action="store", type=int,
+		default=21, dest="port", help="port to listen on"
+		)
+	parser.add_argument(
+		"-u", "--user", action="store", default=None, dest="user",
+		help="username (default: anonymous)"
+		)
+	parser.add_argument(
+		"--pswd", action="store", default=None, dest="pswd",
+		help="password"
+		)
+	parser.add_argument(
+		"--perm", action="store", default="elradfmwM", dest="perm",
+		help="permissions of the user"
+		)
+	parser.add_argument(
+		"--path", action="store", default=os.getcwd(), dest="path",
+		help="path to serve"
+		)
+	ns = parser.parse_args()
+	if (ns.user is not None) and (ns.pswd is None):
+		print(
+			_stash.text_color(
+				"Error: If user is given, pswd must also be given!", "red"
+				)
+			)
+		sys.exit(1)
+	if (ns.pswd is not None) and (ns.user is None):
+		print(
+			_stash.text_color(
+				"Error: If pswd is given, user must also be given!", "red"
+				)
+			)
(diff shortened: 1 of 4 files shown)


ywangd:
I already had monkey patch applied to the i/o streams, see https://github.com/ywangd/stash/blob/master/system/shiowrapper.py

Though normally I am not a fan of monkey patch, it is somehow inevitable and necessary in a confined environment like Pythonista. So it sounds alright to me to monkey patch things like `os.popen` etc, as you said they don't work before the patch anyway. Instead of a command to enable/disable the patches, it could be an configuration option that can be set in the stash config file (i.e. `.stash_config`), with default to be `disabled` (e.g. https://github.com/ywangd/stash/blob/master/stash.py#L52) and users can opt-in to enable it in their own config files.

bennr01:
@ywangd Thanks for your response. While it would be more logical to make the monkey patching a configuration option in `.stash_config`, i think a command would be more flexible and a less hidden feature.
Some reasons why i think a command is more flexible:
   - Monkeypatches could be enabled and disabled without restarting StaSh
   - A command could do more than just enable and disable the monkeypatches, for example showing enabled monkeypatches
   - A command would be less hidden (StaSh has a lot of awesome features which are not easy to be found)
   - I planned on making the patches in a way they could be enabled/disabled individually. So a user could write something like `monkeylord enable os_popen`, `monkeylord disable ALL` or `monkeylord enable INSTABLE` (where ALL and INSTABLE would be groups). Unless we would create a whole section in `.stash_config`, this would become pretty messy in the config.

I will first try to write a few monkeypatches and later check both ways.

My current plan is to create a dir `/system/patches/` with a subdir `/system/patches/modules/`. `/system/patches/` would also be a package containing all monkeypatches as classes. `/system/patches/modules/` would contain modules not implemented in Pythonista, which would be enabled/disabled by modifying `sys.path`.

Is there a way of getting return/exit-codes from `_stash()`/`_stash.run()`? I tought about using `os.getenv("?")`, but with `persisten_level=2` the exit-codes would not be available, are they?
Thanks.

ywangd:
@bennr01 OK if it is more than just a single boolean option, a command is more appropriate. 

Where is the directory `/lib/patches/` created? You won't be able to write against the actual root directory on iOS. Is the command name actually moneky**lord**? I thought it was monkey**load** from the first look of it and it made sense. But monkey**lord** is definitely much more interesting and humorous. 

Talking about hidden features, I have a few ideas to make StaSh more user friendly. First of them is to create a settings menu, similar to what Pythonista itself provides. So current config options and more can be specified via a GUI instead of going to the text file. However I am currently busy with other things and probably won't be able to work on it for quite some time.

It is so incredible that you read the code, noticed the `persistent_level` parameter and even how it is used. Sometimes I myself get confused about all the sub-shell business. This part of code took me quite a long time to sort out and it still not as polished as I had hoped for. Anyway, you can check the exit code of the call interface like the following:
```python
worker = _stash('some_command ...')
print worker.return_value  # this is the exit code
```

bennr01:
@ywangd Thanks for the help with `worker.return_value`. I overlooked it, but instead found out about `worker.state.environ`.

I wanted to create the dir `/patches/` in the `/$STASH_ROOT/lib/` dir, not in the iOS root dir, otherwise `selfupdate` would not update the directory. Sorry if i wrote this in a confusing way.

I am not sure about the name yet. Some of the reasons I tought about using `monkeylord`:
   - A `monkeylord` would be someone commanding monkeys
   - It is more funny
   - The "monkeylord"  was a powerful unit in the strategy-game "Supreme Commander". While this is no real reason for naming a command like this, I tought someone might see it and laugh.

However, the name `monkeyload` would make more sense.
Should I use `monkeyloard` as a compromise?

There is another problem about the monkeypatching: Python2 and Python3 cross compatibility. I tought about this when you created the `universal` branch. You are probably trying to make a Py2 and Py3 compatible version of StaSh like discussed in #179 (at least the name "universal" suggests this). I am not sure how monkeypatching will work with two python versions. Python2 and Python3 have some different functions (`os.popen` is deprecated in python3) and like @dgelessus said in https://github.com/ywangd/stash/commit/de96ff0feccce2fde168cfc7f21c4d3726410085, some functions expect other argument-types in python3. In this case, maybe a config-option would make more sense than a command, because the monkeypatching would then be handled by the StaSh-core.

Talking about monkey-patching, should i try to patch `os.getpid` to return `ShBaseThread.job_id`? This may be usefull for monkeypatching some other functions like `os.kill`.

If you are busy i may be able to write a GUI for the settings. What kind of GUI would you like to add? A ´ui.View`/`dialogs`, some text-UI or a texteditor like vim mentioned in #118? A `ui.View` would be more user-friendly, but a text-ui would keep the terminal/shell feeling of StaSh. Should the GUI replace the command `stash_conf`?

ywangd:
* I think `monkeylord` is good
* Python3 compatibility is still a long way ahead. So you probably don't need to worry about it for the time being. In worst-case, the mokeylord command can check the interpreter version and report it cannot function if the interpreter is Python 3.
* I am not sure about patching `os.getpid` as it currently does have a meaningful behaviour. I guess it really comes down to what you want to do with the patch. If the benefits is high, maybe it is OK.
* The settings GUI would be something similar to the Pythonista's settings dialog, not a text UI. Otherwise might as well just use `stashconf`. Yes it is meant to be a replacement of `stashconf` and more, things like changing the color scheme, button to add editor action for `launch_stash`. I'll probably add a stub implementation in the next a few days. So we have something more solid to talk about.

PR #185, opened 2016-06-22
man.py: support for non-command help

bennr01:
The `man` command can now show general help (=help for non-commands).

New `man` command usage:
   - if no arguments are given, show all commands
   - if one argument is given:
      1. check if a command with that name exists in `$STASH_ROOT/bin/`:
          - If it does exists, print the docstring
          - Otherwis, print an error message and exit
      2. otherwise, check wether a dir or a file with that name exists in `$STASH_ROOT/man/`.
         - if a dir  exists with that name:
             - if `(` and `)` are in the name (they are ignored in the check for a file in `$STASH_ROOT/man/`), read the number between them, otherwise set that number to one, then repeat step 2 with a file named `page_NUMBER.XXX` from this dir.
         - if a file with that name exists:
               1. check file extension
               2. if file extension is `.txt` or unknown, print content (but pause after 100 lines)
               3. if file extension is `.html`, open the file in `quicklook`
               4. if file extension is `.url`:
                    - if file content startswith `stash://`, repeat step 2 with a file named like the rest of the content in `$STASH_ROOT`
                   - otherwise, open the address in the content in `webviewer`

I already added a few help pages:
    - `.url` files redirecting to the `CHANGES.md`, `LICENSE` and `readme.md` files
    - a `.url` file opening the git-tutorial @jsbain made for StaSh
    - a `.txt` file containing the simplified shell grammar from `system/shparser.py`

PR #198, opened 2016-07-27
Monkeypatch-manager

bennr01:
This is going to end up as a long description...

**Short summary**:
This pull requests adds the `monkeylord` command used for managing monkeypatches as discussed in https://github.com/ywangd/stash/pull/190.

**New Command**:
`monkeylord`: simple monkeypatch manager.
Usage: `monkeylord [-h] {enable, disable, list} [name]`.
- `monkeylord -h` will show some help generated by `argparse`.
- `monkeylord list` will show a list of all available patches and wether they are enabled or not.
- `monkeylord enable [name]` will enable the patch with `name`. The default for "name" is `STABLE`.
- `monkeylord disable [name]` will disable the patch with `name`. The default for "name" is `ALL`.

The `name` argument is ignored for `monkeylord list`.

**New help/documentation**:
Type `man monkeypatching` to get more help about `monkeylord` and available patches.
The `monkeypatching` uses the folder-system added in https://github.com/ywangd/stash/pull/185, meaning that `man monkeypatching` will only show the first page. To see a specific page, type `man monkeypatching(pagenumber)`.
(I just noticed that this is not mentioned anywhere except in these pull-requests.)

**Changes to the StaSh-Core**:
- `ShRuntime.run()` now accept two additional optional arguments: `environ` and `cwd`. These are passed to the `ShThread`s.
- `ShBaseThread` now accept two additional optional arguments: `environ` and `cwd`. The corresponding `ShState` will be updated with this values.
- `ShBaseThread` now has a `on_kill()` functions which will be called by the subclasses. This is used to keep track which thread killed this thread. The `job_id` of the killing thread is then saved into `self.killer`. Some monkeypatches require this data.
- `$?` is now set to `127` if a command was not found.
- There was a additional optional argument to `StaSh.__call__()`, which i removed later.
I am a bit worried that i may have broken something, but the tests seem to run fine (at least on travis-ci, they failed on my iOS-device with "cwd changed". However, the local tests also failed for a fresh installation from `ywangd/master` and still used `unittest`).

**Updated `libcompleter`**:
I have added autocompletion-support for the `monkeylord`-command (but only for the first argument).

**Implementation details**:
- The monkeypatches are saved in the package `$STASH_ROOT/lib/mlpatches/`.
- All Patches are subclasses of some baseclasses defined in `$STASH_ROOT/lib/mlpatches/base.py`.
- All Patches are registered as instances in a the `PATCH`-dictionary defined in `$STASH_ROOT/lib/mlpatches/patches.py`.
- There are currently 5 Types of Patch-baseclasses defined in `base.py`:
   - `BasePatch`: baseclass for all other Patches. This class provides Py-version controll and state(enabled/disabled)-controll.
   - `FunctionPatch`: this class is used for patches which replace a module attribute. A subclass needs to overwrite `module`, `function` and `replacement` (=4/5 lines of code to define a new patch). If `replacement` is a function,  the patch will be passed as the first argument.
   - `ModulePatch`: this class is used for patches which add or replace a whole module. Subclasses need to overwrite `relpath` and `name` (=3/4 lines of code to define a new patch).
   - `PatchGroup`: this class is used to define a Group of patches. Subclasses need to overwrite `patches` (=2/3 lines of code to define a new patch).
   - `VariablePatchGroup`: like a `PatchGroup`, but instead of subclassing and overwriting `patches`, it can be used by simple passing the patchlist in `__init__()`. This is usefull for runtime-defined PatchGroups.
- The recommended way to define a new patch is to do something like `class Mypatch(base.PatchGroup):\n   patches=[p1, p2, p3]`.

**Available PatchGroups**:
- `ALL`: All Patches.
- `STABLE`: All Patches defined in `mlpatches.patches.STABLE_PATCHES`.
- `INSTABLE`: All Patches defined in `mlpatches.patches.INSTABLE_PATCHES`.
- `OS`: All Patches for the `os`-module.
- `OS_POPEN`: The patches for the `os.popen*` functions.
- `OS_PROCESSING`: Patches for making the Thread-system behave a bit more like real processes.

**Available ModulePatches**:
- `subprocess`
- `popen2`

**Available FunctionPatches**:
- `os_popen`: Patch for `os.popen()`.
- `os_popen2`: Patch for `os.popen2()`.
- `os_popen3`: Patch for `os.popen3()`.
- `os_popen4`: Patch for `os.popen4()`.
- `os_system`: Patch for `os.system()`.
- `os_getpid`: make `os.getpid()` return the current workers `job_id`.
- `os_getppid`: make `os.getppid()` return the parents `job_id`.
- `os_kill`: make `os.kill()` use StaShs `kill`-command.

**Problems and Informations**:
There a currently some implementation problems:
   - `subprocess.Popen()` ignores a lot of arguments.
   - `subprocess.Popen.terminate()` == `send_signal()` == `kill()`
   - Many user-scripts expect the I/O to be closed once a command finished, but i dont know how StaSh behaves.
   - User-scripts may expect different behavior of some commands (e.g. `ls /does/not/exists` should set `$?` to 1 instead of 0).
   - I could find out wether `os.popen("some cmd").close()` should block.
   - The Python documentation does not specify wether returncodes in `popen2` should be in "normal" or in the "os.wait()" format. I decided to use the "normal"(=exitcode).
   - There may be some other behavior mismatches.
   - The worker created by some `popen`/`system`-functions can be seen in the `jobs`-command, but `fg <job_id>` will raise an `AssertionError`.
   - Documentation has been copied from the Python2 documentation.
   - The way `monkeylord` is implemented relies heavily on the shared module cache. Changes to the import-mechanism or the way scripts are executed may break the patches.
   - most patches can be enabled/disabled without restarting StaSh or Pythonista.
   - The changes can be tested using `selfupdate -f bennr01:monkeylord`.
   - I think i forgot something...

Thnaks for the good documentation and you feedback in https://github.com/ywangd/stash/pull/190 .
**EDIT**: removed the info about the `os.kill()`-bug. There was None.


ywangd:
* You added a `wait` parameter to stash callable interface at a time but then it was removed after merging. Is this intended or by error?

* It seems to me that the new parameters `environ` and `cwd` is a way to inject additional states for a sub-shell execution. Is it right?

* By looking at the code, all jobs created by `popen` is in foreground already (`is_background=False` in the `run(...)` call). So using `fg` on them does not make sense to me. Can you given an example?

* Do you mean thread kill is not possible from another thread except on the UI thread. I didn't know that. Maybe you can try decorate the your call with `on_main_thread`, e.g. add `@on_main_thread` for `Popen`'s `terminate` method?

Still need a lot more time to read on all the changes. Thanks for making this massive pull request.

bennr01:
@ywangd Sorry, some of my notes may be misleading (you probably noticed my terrible english skills).

- The `wait`-parameter was removed intentional. I rechecked the code a few commits after d855af76697f294373f2c25effb1ac29c5e377f0 and noticed that the way the `wait` parameter was added may break other scripts. So i changed the code to use `_stash.runtime.run()`.
- Yes, `environ` and `cwd` are used to inject additonal states. This is used by the `subprocess`-replacement. (see https://docs.python.org/2/library/subprocess.html#subprocess.Popen for more details) 
-  It is a bug that jobs created by `popen* `-functions can be seen using `jobs`. Do you know any way of fixing this? I mentioned the `fg`-command because the bug is noticeable by using `fg` on the created job, which will raise an `AssertionError`. Sorry about the misleading formulation.
- Again, a formulation mistake caused by my terrible english skills. Sorry. I meant the pseudo-mainthread of a script/command (=Any ShBaseThread started by the ShRuntime). Some monkeypatches would be contra-procuctive if they would return the "real" PID, so they will instead return the "job_id".  The `PROCESSING`-Patches contain some functions for consistency, for example a `os.getpid()`-replacement. The replacement returns `threading.current_thread().job_id`, but this only works when called on a `ShBaseThread` (or the subclasses).
I just noticed some other formulation problems in the pull-request.I am going to edit it soon to fix my mistakes.

PR #201, opened 2016-08-05
Added ifconfig.py

zhunsaker:
For now the implementation is very minimal, I am currently experimenting, but hoping others can help get this a little further than it is right now. I want to be able to show local and public IP's and possibly hardware information, it may require the use of the OBJC bridge.


fschaeck:
Please create pull-request against branch dev of this repository. Thanks!

PR #202, opened 2016-08-10
modify git.py to pass ssh_keyfile to ssh client -> ssh connection to github works again

RichardPotthoff:
I am not sure what caused it, but when I recently tried to use StaSh's "git push" command, I got the error message:

 stash: \<class 'paramiko.ssh_exception.AuthenticationException'\>: Authentication failed.
 
My patch fixed the problem for me.

### DIFF
--- a/bin/git.py
+++ b/bin/git.py
@@ -128,9 +128,35 @@ else:
     from dulwich import porcelain
     from dulwich.index import index_entry_from_stat
     from gittle import Gittle
-
-dulwich.client.get_ssh_vendor = dulwich.client.ParamikoSSHVendor
-#  end temporary
+    
+# create wrapper class to pass ssh_keyfile to ParamikoSSH client
+try:#... 
+    def find_ssh_keys():# "borrowed" from ssh.py, where it worked
+        ssh_dir = os.path.join(os.environ['STASH_ROOT'], '.ssh')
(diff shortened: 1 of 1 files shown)


jsbain:
Did this work for you previously?  Did you create a new key recently, or delete the keys in ~/.ssh as well?  What paramiko version are you using?
The git usage of paramiko depended on keys in ~/.ssh.  ssh_keygen creates keys in both locations.  it should not really be necessary to create a custom wrapper, if the keys are in the default paramiko location...

RichardPotthoff:
That explains it!
The home directory changed when I upgraded to Pythonista 3. That's why "git" could not find the keys anymore. 
However, the "ssh" command did find the keys in the STASH_HOME directory. So I tried to get the "git" command to behave in the same way as the "ssh" command, without having to modify "gittle" and "dulwich". 
I did not create new keys, because I would have had to distribute the new public key to a lot of different places. I knew that the old keys still worked, because the "ssh" command still worked.
P.S.: I just checked my Paramiko version: It is version 1.16.0

ywangd:
@jsbain I don't quite follow the conversation here. So is this change necessary and need to be merged? Thanks!

RichardPotthoff:
The merge is not necessary. Putting the private key in ~/.ssh fixes the problem without the patch to git.py.

PR #207, opened 2016-08-23
added easy_config

bennr01:
See Discussion in #190.
Unfortunately, i have been only able to test this on a 1136 × 640 Display.
However, this script does not use any "magic numbers", so it should work fine on all screen sizes.
Maybe someone can test this?
I was also not able to find a way to automatically add an editor shortcut.
This does not (yet) replace `stashconf`, because `stashconf` is a simple way to temporary change the config. Should i rename `easy_config` to replace `stashconf`?
Also, the `[display] BACKGROUND_COLOR`-option seems to be broken. The background-color is set correctly whenever StaSh is started, but whenever new text is printed, it will have a black background-color.
**EDIT**: Why does the CI-Build fail? I have not changed anything in the threading-system in this commit and the command i have added is not even being tested in the tests. Did i broke something in some previous commits?

### DIFF
--- a/bin/easy_config.py
+++ b/bin/easy_config.py
@@ -14,16 +14,29 @@ ORIENTATIONS = ("landscape", "landscape_left", "landscape_right")
 TYPE_BOOL = 1
 TYPE_INT = 2
 TYPE_STR = 3
-TYPE_FILE = 4
+TYPE_FILE = 4  # NotImplemented
 TYPE_COLOR = 5
 TYPE_CHOICE = 6
 TYPE_LABEL = 7
+TYPE_COMMAND = 9
 
 CONFIG_PATH = os.path.join(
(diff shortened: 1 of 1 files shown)


ywangd:
I think the CI build sometimes (rarely) fails due to the somewhat badly written test cases for multiple background jobs (my fault). Something to fix later ...

The background color issue is due to the use of attributed text. Each character can have its own background color which is by default set to black (code in `shscreens.py`). It would be great if all the colors used for them are configurable as well. But that's quite a big task that requires some level of re-design for the color systems. 

Otherwise, the new script works great. Thanks a lot for your effort.

ywangd:
@jsbain developed an editor action script and is available here https://gist.github.com/jsbain/c9f42c81c53b276b6560

bennr01:
@ywangd thanks for the feedback and the link. i will try to add this and then open another pull-request. Thanks!

PR #210, opened 2016-08-27
easy_config.py: added "Add Editor Shortcut"

bennr01:
The `easy_config`-command now provides a button to add a editor-shortcut for `launch_stash.py`.
This script requires [this module](https://gist.github.com/jsbain/c9f42c81c53b276b6560), which is automatically downloaded when the button is pressed for the first time. It is not downloaded when the command is run the first time to improve off-line capabilities.

@jsbain would you mind if i later include your `add_action.py`-module directly in `$STASH_ROOT/lib/`?

### DIFF
--- a/bin/easy_config.py
+++ b/bin/easy_config.py
@@ -2,13 +2,19 @@
 import os
 import ast
 import threading
+import imp
 import console
 import ui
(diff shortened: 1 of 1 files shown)


jsbain:
i don't mind.

PR #212, opened 2016-09-23
New beta

jsbain:


jsbain:
This corrects issues with the new beta -- subprocess is missing Popen and call (apparently dulwich was calling those, which is surprising) Pythonapi location changed, and outputcapture befame pykit_io.

Issue #167, opened 2016-03-05
[idea] mount command


jsbain:
I am not sure how this would work exactly, but seeking ideas.

it would be convienent to have a "mount" type command for attaching external file systems into stash.  Think dropbox, google drive, github, webdav, samba, etc...   We would create a set of "device drivers" which implement common file and folder ops, such as open, read, write, listdir, rmdir, mkdir, mv, cp, fstat, and a few others.  Then after mounting, stash would keep track of mounted drives, and any file ops using those paths would get farmed out to the "driver".   Ideally it would be neat if any script that uses the os or shutil libraries somehow would work out of the box, but that probably requires great feats of monkey patching which maybe we don't want.  An alternative would be to rewrite all supported commands in stash to make use of special stash "devices".

thoughts?

ywangd:
Yes ideally it would be great if the `mount` command can work like its true counterpart so that other commands can work without changes. But it is probably very difficult if not impossible. Two possibilities are:

1. Using `ctypes`, it seems that we have access to libc's mount function (`ctypes.CDLL(None).mount`). But I haven't tested it and not sure if it is possible with iOS. However, even this allows mount of some NSF drive or even Samba, I don't see how Github repo can be mounted (which is likely the main use case for this command).

2. Monkey patch relevant Python libraries. I would not recommend this approach as it is difficult and result is likely to be suboptimal. There are some cases that the patch just won't work. For an example, if a module is imported already before stash runs and it has an import like `from os import listdir`, the monkey patch afterwards won't be able to help here as the module saves its own copy of listdir. Thus this module remains unaware of the "device drivers". There may be some convoluted way to fix this with the `gc` module (search for `listdir`'s reference and modify them). But it feels too hacky and unstable to me. (I tried something along this direction when working on allowing parallel running scripts each has its own copy of `sys.argv`. But the result was always a mess).

If we cannot take any of above approaches, we are not able to make "native" device drivers. So we'll have to either:

1. Rewrite all relevant stash commands. I am a bit hesitate about this option. This requires not only rewriting of many commands but also stash now has to actively tracking the status of all mounted devices. 

2. Create specific command for each type of device drivers, e.g. `gh` can be used for github repos. So you use it like `gh mount owner/repo`. Or you can have `gh` enters an interactive mode after mounting. Something like the FTP sesssion, for an example:

```bash
[stash]$ gh mount owner/repo
mounting owner repo ..
[(gh)owner/repo]$ ls
remote_dir/
[...]

PR #213, opened 2016-10-10
added "mount" and "umount"; improvements and bugfixes

bennr01:
**Short summary**:
This pull request adds a mounting system using the monkeypatching approach (but without thread-local modules; the results were to unstable). It implements #167.

**New commands**:
**`mount`**:
   This commands allows the mounting of filesystems. Most of the filesystem code was already implemented in the `mc`-command. Supported filesystem types: `dropbox`, `ftp`, `zip`, `local`. The documentation as usage can be viewed using `man mounting`. Monkeypatches are managed using the `monkeylord` command and are automatically enabled when `mount` mounts a filesystem. The user is warned before any monkeypatches are enabled (except when `-y` was passed to `mount`). Unfortunally, not every command was fully compatible with the monkeypatches. The changes are mentioned later in this pull-request.
**`umount`**:
   This Command unmounts a filesystem. I have modified `shruntime.py` to automatically add an alias `unmount="umount"`. See `man mounting` for more details.
**`dropbox_setup`**:
   This command is used for configuring the dropbox. The dropbox setup was once defined in `mc`, but has been splitted into a new command. Scripts can access configured dropbox account using the new `stashutils.dbutils`-module.

**Compatibility changes**:
Unfortunally, not everything was compatible. Here are the changes i had to make:
   - `shruntime.py`: `execfile()` does not support custom `open` replacements. This caused some problems when the CWD was inside a mounted directory. I replaced `execfile()` with a mixture of `open`, `compile`, and `exec` (using `runpy` caused problems when used with `ui`-scripts). It seems like this improved the performance a bit (I havent timed it, but it feels a bit faster). I have not noticed any problems caused by this.
   - `ls.py`: `ls` prints the color based on the filetype. Unfortunally, `tarfile` does not support custom `open`-replacements, which caused `ls` to raise Exceptions. I have added a `try: ... except: ...`-statements, which will just assume it is not a tarfile when it cant open the file using `tarfile`. Images and zips are still colored.
   - `test_runtime.py`: This test compares the output of `alias` with a hardcoded string. Because i have added a alias `unmount` for `umount`, this test failed. I have now added the new alias to the string.

**Other Changes**:
   - there is a new package `$STASH_ROOT/lib/stashutils/` for shared functions.
   - most of the code of the `mc` command was splitted out into seperate modules.
   - The `pythonista_add_action` module is now included and wont be downloaded. Thanks to @jsbain for sharing his code.
   - `ls.py`: now exits with status 1 if an error occured.
   - `webviewer.py`: added foreground option.
   - The `DropboxFSI` now uses the dropbox APIv2. This may cause compatibility problems with earlier dropbox configurations created by the old `mc`-command. In this case, you can use `dropbox_setup` to remove the old configuration.
- probably some changes i forgot about.

**About perfomance on mounted filesystems and FSI-bugs**:
The performance on mounted filesystems is pretty bad. StaSh does a surprisingly high number of I/O operations for each command. When searching for a command, StaSh calls 4 times `os.path.exists()` before executing a normal command. `ls` opens every file in the directory to list several times to check its filetype. I have added some optimizations to some FSIs to improve the performance. For example, the `DropboxFSI` will only read as much as required from the remote file and buffers it in RAM.
Some FSIs(=FileSystemInterfaces) have unique bugs. For example, the FTP-FSI does not work with every FTP-Server (it requires the `size`-command) and sometimes gets connection-timeouts. The Zipfile-FSI does sometimes behave weirdly with subdirectories (I have no idea why).

Sorry for my terrible english.


ywangd:
wow!! You are bringing stash to a whole new level! Thanks a lot for your effort. 

The stash internals still have much room to optimise. I didn't pay too much attention to performance when writing the code. Now with the remote file system, we have a different situation. 

Also I thought about restructure/splitting the stash core even further to help with performance as well as overall code/project management. Though stash is envisioned as shell for pythonista, it actually is a lot more than just a stripped down shell. From what I can see, it really provides a **shell**, a **terminal emulator** and a **mini operating system**. So it seems to make sense if the code is also organised that way. What do you think?

PS: Your English reads perfectly fine to me. :)

bennr01:
@ywangd I dont think that it the performance is a big problem. Most of the performance issues can only be noticed when using the slow monkeypatched functions. Most of the I/O calls are required. However, the `ls` command could be optimized to check if it is running on a remote filesystem (using `mlpatches.mount_ctrl.get_manager().get_fsi(path)[0]!=None`) and if it is, then determine the filetype only by checking the file extension. Most of the other commands are only occasionally used or dont do any unnecessary I/O calls.
The StaSh core itself is pretty well structured (i remember the time when it was a single file). However, have you tought about splitting the shell (but not the parsing) into a separate command, which would be called when the prompt would be shown? I once tried to implement it, but ran into issues with `shexpander` and `shparser`(duplicated input, ...). This would be useful in conjunction with `StaSh.__call__` and custom stdins/stdouts (for example, using a remote shell access). It may also be useful if we could extend the sh parsing to parse more complex sh-scripts (unfortunately, i am pretty useless for this task because python is my only programming language and i dont understand regular expressions). I also think that it may be useful to define permanent aliases without modifying `.stashrc` (This would be useful for scripts and commands, for example `pip`).
(A short off-topic question: Would it be possible to add a `screen`-command? I know we can use `ctrl+Z` to push a job to the background, but it is kind of annoying not to see the last output).
If i remember correctly, github added a way to manage such projects better (https://github.com/universe-2016#projects). Maybe you should create some projects for this tasks?

PR #216, opened 2016-10-31
Ability to pass command using an URL; small changes

bennr01:
This pull-requests contains three small changes:
- **Run a StaSh command when clicking an URL (using the pythonista URL-scheme):** `launch_stash.py` now accepts a command as optional argument, which will then be run instead of the `totd`-command when StaSh is started. This allows the use of StaSh as an utility for easily sharing code, support other users or run installs. I have created a small [script](https://gist.github.com/bennr01/789c0a6e27a0a458f615dc4464c93030) which helps to encode a StaSh command into a URL. Also, the StaSh-class now accepts `command` as an optional argument. If command is None, the default command (`totd`)will be run on startup. If command is False, no command will be run on startup (This does not show the prompt when StaSh is started because `script_will_end` will not be called, but is required if you want to use `_stash(cmd)` after initialization.). If command is a string, it will be interpreted as a command. Example: [pythonista3://launch_stash.py?action=run&args=%22selfupdate%20-f%20bennr01%3Arestructure%22]() installs `pytest` from pypi. I am a bit worried someone might use this in combination with the `rm -r $HOME2` command and disguises the link, but hopefully no one does this.
- **optimized `ls.py`:** like discussed in #213, `ls.py` now checks wether the path it should list is a mounted path. If it is, it will only check the file-extensions to determine the filetype, otherwise it behaves normally.
- **moved `mount_ctrl.py`:** I think it makes sense to move this file from `$STASH_ROOT/lib/mlpatches/` (The path where the monkeypatches are saved) to `$STASH_ROOT/lib/stashutils/`(The path where the mounting system except the monkeypatches are saved).


bennr01:
@ywangd Sorry for being a bit impatient, but i have two questions:
1. Should i change some commits or is there any reason this pull-request has not yet been merged? If there is anything wrong with it, just say it and i will change it (I think the most likely reason is that you are/were busy and just did not see this pull-request, which is why i am writing this message).
2. About the reorganization of the code you mentioned in #213: Are there any new plans? I would like to add a directory for extensions in the documents-folder, which could then be used for command installed by scripts using `pip`, easily adding custom monkeypatches and for adding FSIs (the abstract interface the `mc`, `mount` and `umount` command use). Are you ok with this?
Thanks

ywangd:
Somehow I thought I have merged this pull request. But apparently I didn't. Sorry for the unnecessary delayed action. 

I haven't got any new plans for the code yet. Honestly I haven't been working on it for a while now due to lack of time. If you are still interested in improving it, I could add you as a collaborator of this project so that you have full control of this repository. Is this something you would be interested in? @jsbain Are you OK with being a collaborator as well?

bennr01:
@ywangd dont worry about it. Everyone forgets something from time to time.
Thanks for inviting me to become a collaborator. I am interested in becomming a collaborator, but i do not think it would be necessary. As long as you accept (or deny) my pull-requests every now and then. But thanks for the offer and the answer.

Issue #160, opened 2016-02-19
`help` will print *.py files under ~/Documents/bin


wizardofozzie:
I'm not great with debugging, but I'll do my best here.

Basically, I had a directory structure as so:

`~/Documents/bin/remove_metadata.py`
`~/Documents/bin/noop.py`

Then, when I go into stash and type `help`, the help messages would print up until `rm`, but it would also list `noop` in the help commands.

So basically, stash is reading the wrong `bin` directory (instead of `~/Documents/site-packages/stash/bin`, it's also reading `~/Documents/bin`

I solved the issue by moving any `*.py` files from under `~/Documents/bin`

[Error screen paste](https://gist.github.com/anonymous/7972c6f56bc7570b1c8e)

ywangd:
Both `bin` directories are by default the locations where StaSh searches for command scripts. This is intentional as it is recommended to place user scripts under `~/Documents/bin/` so they do not interfere StaSh's internal `bin` (`~/Document/site-packages/stash/bin/`) but can still be called from StaSh's console.

With above said, you are right that the `help` command should only look into the internal `bin` folder. It can only hope sensible help messages to be generated for builtin commands NOT user scripts. It should not assume user scripts to follow a certain format or even complete. I'll make the change.

Thanks!

wizardofozzie:
Yeah, in the cases where there are no help docs there's no issue with the parsing, since it lists the file name (command) only.
Thanks again!

ywangd:
This issue is fixed with commit 29ff695934719e03178ead05e60e2d558e873924
It is currently available in the `dev` branch and will be merged to `master` soon.

PR #226, opened 2016-12-14
Made StaSh more extensible; pip now handles command line scripts; fixes

### LINKED ISSUE #167: [idea] mount command

### LINKED ISSUE #225: Launch via wrench menu executes open editor file
Pythonista supplies the current editor tab's path as an argument to any command chosen in the wrench menu. This is great if you have e.g. a word counting utility, because that will operate on the file open in the editor.
However, launch_stash executes any command specified as an argument on the command line instead of totd (as of latest change of 23 Oct). This is fine if you start stash by opening launch_stash and selecting run; no arg, no harm done. But if you start stash from the wrench menu (as facilitated by easy_config), it executes the file open in the editor prior to the stash repl. This is usually **NOT** what you want.

bennr01:
This pull request makes StaSh more extensible, extends `pip` and contains some fixes.

When StaSh launches, it will now create some folders for extensions:
- `$HOME2/stash_extensions/`: parent directory for extensions (to keep them sorted).
- `$HOME2/stash_extensions/bin/`: This directory is now in `$BIN_PATH`. It may/can/should be used by scripts automatically creating StaSh commands. While it would also be possible to create the commands in `$STASH_ROOT/bin` or `$HOME2/bin`, both have some limitations. `$STASH_ROOT/bin` may be overwritten during an update (or worse, some command may overwrite an important StaSh command). `$HOME2/bin` may contain scripts by the user or other data, which means that no script can be sure it is allowed to overwrite the contents of this directory. Also, due to #160, `man` wont print docstrings for commands in this directory.
- `$HOME2/stash_extensions/man/`: the `man` command will also look here for non-command topics.
- `$HOME2/stash_extensions/fsi/`: the `mount` and `mc` commands will try to load FSIs (=FileSystemInterfaces, the abstract class used by the mounting system to interact with filesystems) from modules in this directory. It will import these files and check for a dictionary named `FSIS` (mapping the names to the FSI-classes) and then register them.
- `$HOME2/stash_extensions/patches`: the `monkeylord` command will look for files in this directory, import them and look for variables named `STABLE_PATCHES` and `INSTABLE_PATCHES` (both mapping the name of the patches to patch instances) and then register them.

However, most of these checks are only done whenever they are first required and thus may require you to restart Pythonista before using the extensions.
The directory in which the extensions are stored (default: `$HOME2/stash_extensions/`) can be changed by changing a variable in `shcommon.py`.
There is also an API to create these extensions available as `stashutils.extensions` (stored in `$STASH_ROOT/lib/`). The API automatically creates the file in the correct location (which makes changing the installation directory easier because you would only need to change one line).
The corresponding documentations (`mounting`, `monkeypatching`...) have been updated.

I have changed the `command` argument added in #216 into a `--command COMMAND` argument and make `launch_stash.py` ignore additional arguments to fix some problems in the beta. This should fix #225, however i am unable to test this due to not having access to the beta.

The `man` command can now list all non-command topics by using `man topics` and will ignore invalid python files.

And last, but not least (and probably the most important of this pull request): The `pip` command now handles command line scripts defined in the `setup()` call.
It handles both the `scripts` keyword argument and the `entry_points` keyword argument, however it will only create the docstrings for `entry_points={"console_scripts":...}` commands (because the docstring of files using `scripts` may define their own). Also, only the `console_scripts` entry point will be handled and other entry points will print a warning.
The commands are created using the extension API defined in `$STASH_ROOT/lib/stashutils/extensions/` and thus create the commands in `$HOME2/stash_extensions/bin/`.
For example, `pip install django` will install the `django-admin` command (this may require a restart of Pythonista, but this is varying between different packages). `pip install kvndb` creates the `kvndb` command and the `twistd`(due to having `twisted` as a requirement)`.
This greatly improves the extensibility of StaSh due to making StaSh capable of using the ten-thousands of command line interfaces available on pypi.

By the way, here are a few questions (you dont need to answer them if you are busy):
1. Should we/you close #167 (The `mount` idea issue)? The `mount` system is now mostly implemented.
2. About python2/3 cross-version running: Is the code for executing a statement in python3 from pyhon2 available somewhere?

Thanks

### DIFF
--- a/bin/man.py
+++ b/bin/man.py
@@ -114,9 +114,12 @@ def find_command(cmd):
 
 		
 def get_docstring(filename):
-	with open(filename) as f:
-		tree = ast.parse(f.read(), os.path.basename(filename))
-	return ast.get_docstring(tree)
+	try:
+		with open(filename) as f:
+			tree = ast.parse(f.read(), os.path.basename(filename))
+		return ast.get_docstring(tree)
+	except:
+		return "UNKNOWN"
 
 		
 def get_summary(filename):
--- a/bin/man.py
+++ b/bin/man.py
@@ -8,6 +8,8 @@ import ast
 import os
 import sys
 
+from stash.system.shcommon import _STASH_EXTENSION_BIN_PATH, _STASH_EXTENSION_MAN_PATH
+
 if sys.version_info[0] > 2:
 	# py3 compatibility
 	raw_input = input
@@ -18,22 +20,29 @@ TYPE_CMD = "command"
 TYPE_PAGE = "page"
 TYPE_NOTFOUND = "not found"
 
-BINPATH = os.path.join(os.environ["STASH_ROOT"], "bin")
-PAGEPATH = os.path.join(os.environ["STASH_ROOT"], "man")
+MAIN_BINPATH = os.path.join(os.environ["STASH_ROOT"], "bin")
+MAIN_PAGEPATH = os.path.join(os.environ["STASH_ROOT"], "man")
 
-if not os.path.exists(PAGEPATH):
-	os.mkdir(PAGEPATH)
+BINPATHS = [MAIN_BINPATH, _STASH_EXTENSION_BIN_PATH]
+PAGEPATHS = [MAIN_PAGEPATH, _STASH_EXTENSION_MAN_PATH]
+
+for p in BINPATHS + PAGEPATHS:
+	if not os.path.exists(p):
+		os.mkdir(p)
 
 		
 def all_commands():
-	cmds = [
-	fn[:-3] for fn in os.listdir(BINPATH)
-	if fn.endswith(".py")
-	and not fn.startswith(".")
-	and os.path.isfile(os.path.join(BINPATH, fn))
-	]
-	cmds.sort()
-	return cmds
+	all_cmds = []
+	for bp in BINPATHS:
+		cmds = [
+		fn[:-3] for fn in os.listdir(bp)
+		if fn.endswith(".py")
+		and not fn.startswith(".")
+		and os.path.isfile(os.path.join(bp, fn))
+		]
+		all_cmds += cmds
+	all_cmds.sort()
+	return all_cmds
 
 		
 def get_type(search):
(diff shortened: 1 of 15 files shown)


ywangd:
Thanks @bennr01 
* I have now closed issue #167
* As for the python2/3 cross call, I don't have a formally written code. It was mostly messing around with scratch files. I should be able to find it. But not sure if they still working after many new beta versions.

ywangd:
The following code works on the current beta. It calls Python 3 from Python 2.

```python
import ctypes
import os
import sys

EXE_DIR = os.path.dirname(sys.executable)
p2f = 'Py2Kit'
p2 = ctypes.PyDLL(os.path.join(EXE_DIR, 'Frameworks', p2f + '.framework', p2f))
p3 = ctypes.pythonapi

state = p2.PyGILState_Ensure()
p2.PyRun_SimpleString('print 88')
p2.PyGILState_Release(state)

state = p3.PyGILState_Ensure()
p3.PyRun_SimpleString('print(42)')
p3.PyGILState_Release(state)
```

To run it in appstore version, I think you just need change `p2f` to `PythonistaKit`. Note that `p2` and `p3` are independent interpreters and share no state/variables.

dgelessus:
@ywangd `ctypes.pythonapi` might not do what you want, especially under Python 2. Internally, `ctypes.pythonapi` is a `ctypes.PyDLL(None)`, and in the global "`None`" namespace the Python 3 API functions hide the Python 2 ones (I think). However a `PyDLL` only holds the GIL and processes errors in the interpreter from which it is used (in your case Python 2), so using it to call Python 3 functions isn't very useful. You might want to make `p3` a normal `ctypes.CDLL` loading the `Py3Kit`.

If you want to call Python 2 from Python 3, you'd do it the other way around of course - `p2` as a `ctypes.CDLL`, and `p3` as a `ctypes.PyDLL`.

ywangd:
@dgelessus I am not quite sure if I understand what you mean. `ctypes.pytonapi` does load the Python 3 interpreter as this can be checked by execute a simple `import sys; print(sys.version)` string with it. Are you saying it is the Python 3 interpreter but with limited functions like cannot handle errors? But it does report error when I tried a Python 2 style print such as `print 42`. 

Anyway, the biggest issue with the cross interpreter call is that the two interpreters do not share states and also I cannot see a way that allow interactive calls, which are important to StaSh.

bennr01:
@ywangd @dgelessus Thanks for the code. I havent had time to test it yet, but i already had some ideas on how we could execute a python3 code from within python2.
If the biggest problem is the not shared state (especially `sys.std*`), we could simply create some sort of preflight script and use pipes for communication.
The concept:
1. StaSh runs py3 file when executing the command `python3 some_file`
2. Use `os.pipe()` to create 3 (or maybe 4) file-descriptors for pipes: `stdin`, `stdout` and `stderr` (maybe a RPC pipe too?)
4. Command `python3` opens the writing end for the `sys.stdin` pipe and the reading ends for `sys.stdout` and `sys.stderr`.
5. Create some sort of preflight script using a template, then run this preflight script
6. The preflight scripts opens the reading end of the pipe `sys.stdin` and the writing ends for the pipes for `sys.stdout` and `sys.stderr`, then switches CWD and (if required) modify the environment and set `sys.argv`.
7. The preflight script sets `sys.std*` to the pipes
8. The preflight script uses `exec()` to execute the source. (probably in a `try:... finally: ...` statement so it can do some cleanup.
9. Done.
This is only a simplified concept and most of the work will (if it works)  be done by a library (so we can easily use it in multiple places.
This would solve the following problems:
- shared `sys.std*` variables
- shared `sys.argv`
- shared environment
- shared cwd

However, it probably wont work with some commands of StaSh (like `mount` and `monkeylord`, which patch the py2 lib).

What do you think about this?

dgelessus:
@ywangd The definition of `ctypes.pythonapi` is `pythonapi = PyDLL(None)` ([source](https://github.com/python/cpython/blob/master/Lib/ctypes/__init__.py#L432)). `PyDLL` is a specialized version of `CDLL` meant for calling functions in Python's C API. Quoting [the `ctypes` docs](https://docs.python.org/3/library/ctypes.html#ctypes.PyDLL):

> Instances of this class behave like CDLL instances, except that the Python GIL is *not* released during the function call, and after the function execution the Python error flag is checked. If the error flag is set, a Python exception is raised.
> 
> Thus, this is only useful to call Python C api functions directly.

This also only works properly with the C API of the version that your script is running in. If your script is running in Python 2, then a `ctypes.PyDLL` only takes care of Python 2's GIL and error checking, because it doesn't know about Python 3.

In addition to the GIL and exception issues, accessing the Python API via `ctypes.pythonapi` or `ctypes.CDLL(None)` is messy, because a `None` `CDLL`/`PyDLL` looks up functions in *all* loaded libraries/frameworks. So `ctypes.pythonapi` isn't just the Python 3 API, it's anything that happens to be loaded, including some Python 2 API functions (like `PyInt_FromLong`) or other C functions (like `printf`).

Long story short: `ctypes.pythonapi` doesn't work as intended in most cases, and `ctypes.PyDLL` is only useful for calling the current Python's API. For calling "the other" Python, you need to use `ctypes.CDLL` and manage the GIL by hand.

PR #231, opened 2017-03-27
Bugfixes

bennr01:
This pull-request contains these changes and bugfixes:

- `crypt.py` now uses `pyaes` instead of `simplecrypto`. While `pyaes` is a bit slower, it is 100% python. **This change is backwards incompatible!** I changed the AES mode from `CFB` to `CTR`. Due to this, it is no longer possible to decrypt files encrypted with the old `crypt` command. It *would* be possible to make `crypt` backward compatible, but there are also some reasons to do this change. For example, the `CFB` mode requires an initlaization vector (unlike `CTR`). Looking at the sourcecode of `simplecrypto`, this was appended in front of the file. I am not sure wether that would be the most inter-implementation compatible way. There were also some other changes. Maybe we should add a `--compatibility` argument to `crypt`?
- fixed a bug in the monkeypatch for `subprocess`. This was a few months ago, so i am not sure anymore what that bug did.

This puill-request also contains an unfinished version of `librunner.py` (the module for running py3 from py2). I am not sure in which state it is, but unfortunately i dont have the time to continue working on it (at least not now).

### DIFF
--- a/bin/crypt.py
+++ b/bin/crypt.py
@@ -13,48 +13,60 @@ optional arguments:
   -k KEY, --key KEY  Encrypt/Decrypt Key.
   -d, --decrypt      Flag to decrypt.
 '''
-import sys
 import os
+import argparse
+import base64
+
 try:
-    from simplecrypto.key import AesKey
+	import pyaes
 except:
-    print 'Installing Required packages.'
-    _stash('pip install simplecrypto')
-    sys.exit(0)
+	print 'Installing Required packages.'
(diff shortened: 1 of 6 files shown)


ywangd:
Thanks @bennr01 
I personally think it is ok to introduce this backwards incompatibility unless there are serious complaint about it (which I doubt). 

Thanks for your great contribution to stash.

bennr01:
@ywangd Also thanks to you for creating stash. :)

PR #232, opened 2017-04-03
ln for linking and diff for diffing, ls shows sym links in yellow

eddo888:
Hi,

I've created a ln.py bin command for sym linking
a modified ls.py command that shows sym links in yellow and in -l mode with -> as expected.
also included the diff.py command that works like diff using libdiff

Cheers Dave

### DIFF
--- a/bin/diff.py
+++ b/bin/diff.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+
+import sys, os, re, json, argparse, time, pytz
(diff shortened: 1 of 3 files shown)


ywangd:
Thanks for the pull request. 
Could you please change (or recreate) it to target the `dev` branch instead of `master`? Changes in `dev` branch will be merged into master once in a while.

PR #233, opened 2017-04-04
ln for linking and diff for diffing, ls shows sym links in yellow

eddo888:
Hi,

I've created a diff.py command to diff using libdiff,
an ln.py command to do links, and symlinks,
an updated ls.py command to show symlinks as yellow and for ls -la as -> target

Cheers Dave

### DIFF
--- a/bin/diff.py
+++ b/bin/diff.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+
+import sys, os, re, json, argparse, time, pytz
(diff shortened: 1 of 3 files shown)


bennr01:
This PR was merged in https://github.com/ywangd/stash/commit/077a9f026e128390988001c1a149d06fe2f64fc0, but it seems like this PR stayed open. I'll close it now.

PR #257, opened 2017-07-31
Package manager for StaSh (read description please)

Seanld:
Latte is a new package manager built for Latte. It may require updates in the future, so if you're going to accept this, make sure to be willing to accept more requests in the future.


Seanld:
@cclauss It isn't. This was a project I started working on back when I had a working iPad. But now I have a Samsung, and a Chromebook. I can't really fix and debug this, but if you'd like, you can.

Issue #264, opened 2017-08-21
pip doesn't see new package version (caching issue?)


zrzka:
* Did upload new package to PyPI (blackmamba version 0.0.4)
* Did install it via pip, works
* Then I did upload new versions 0.0.5, 0.0.6, 0.0.7 and 0.0.8
* The latest one is [0.0.8](https://pypi.python.org/pypi/blackmamba)
* Issued `pip versions blackmamba` and it shows me just `blackmamba - 0.0.4`
* Did restart Pythonista, same issue, new versions are not visible
* Logged in to PyPI and removed all versions prior to 0.0.8, so, only 0.0.8 is available (see attached screenshot)
* Issued `pip versions blackmamba` and it still shows me just `blackmamba - 0.0.4`
* Issued `pip install blackmamba`, attempt to install 0.0.4 was made and it failed, because it no longer exists

It's more than 3 hours from 0.0.8 release, but still no luck. Is there a way how to force reload PyPI index? Or discard some local caches?

<img width="1146" alt="screen shot 2017-08-21 at 13 22 24" src="https://user-images.githubusercontent.com/1084172/29517204-3e8b5962-8674-11e7-95f0-9a9bca8601d3.png">

zrzka:
Actually ...

```python
#!python2

import xmlrpclib

pypi = xmlrpclib.ServerProxy('https://pypi.python.org/pypi')
hits = pypi.package_releases('blackmamba', True)  # True to show all versions
print(hits)
```

... really prints just `['0.0.4']`.

zrzka:
And also check this output, it seems that the issue is with `True` in `package_releases`.

```python
#!python2

import xmlrpclib

pypi = xmlrpclib.ServerProxy('https://pypi.python.org/pypi')
hits = pypi.package_releases('blackmamba', True)
print('Hits with True {}'.format(hits))
hits = pypi.package_releases('blackmamba', False)
print('Hits with False {}'.format(hits))

urls = pypi.release_urls('blackmamba', '0.0.4')
print('0.0.4 release urls {}'.format(urls))
urls = pypi.release_urls('blackmamba', '0.0.8')
print('0.0.8 release urls {}'.format(urls))
```

And the console output ...

```
Hits with True ['0.0.4']
Hits with False ['0.0.8']
0.0.4 release urls []
[...]

PR #269, opened 2017-09-02
XML-RPC replacement #264

zrzka:
Replaces XML-RPC calls in pip, except package search. Can't find any endpoint and it looks like that people recommends to list simplified index and search in it. Time consuming, not an ideal solution, ... I think that we can live with XML-RPC & search for now. Package is listed immediately and rest is done via JSON.

### DIFF
--- a/bin/pip.py
+++ b/bin/pip.py
@@ -903,6 +903,7 @@ class PyPIRepository(PackageRepository):
         if not r.status_code == requests.codes.ok:
             raise PipError('Failed to fetch package releases')
 
+        # _determin_hit expects latest first, sort them by versions
         tuples = [release.split('.') for release in r.json()['releases'].keys()]
         tuples = sorted(tuples, reverse=True)
         return ['.'.join(release) for release in tuples]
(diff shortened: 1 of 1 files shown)


zrzka:
Ah, it's against master, not dev, will fix it.

zrzka:
Please, do not merge yet. This PR contains quick fix to versions, but I'm going to optimize it now to avoid three requests to the same API endpoint (in `download` for example). I'll let you know when this PR will be ready.

zrzka:
Feel free to merge. What's done in this PR:

* Fixed update
    * Same context manager wrapper as install, because it calls install internally
    * VersionSpecifier has wrong arguments
* XML-RPC replaced
    * In all functions, except search where I'm unable to find any JSON endpoint
* Optimize number of requests by reusing pkg_data (download function for example)

jsbain:
@zrzka It seems you may need to update the tests so travis ci passes?

zrzka:
@jsbain it's failing in test_threads.py. Sometimes passing, sometimes not. I did not investigate these tests because I thought rhat pip has nothing to do with them and someone is working on threads in dev. See the history https://travis-ci.org/ywangd/stash/pull_requests Green, red, green, ... Anyway, will check.

ywangd:
@zrzka You are right. The threads test is not well written and behaves sometimes randomly.

Issue #271, opened 2017-09-11
pip update doesn't ignores -d from pip install


zrzka:
```bash
pip install blackmamba -d ~/Documents/site-packages-3
pip update blackmamba
```

If there's a new version available, `pip update` installs it, but the update is installed in `~/Documents/site-packages` folder, not `site-packages-3`. In other words, before removing package, pip should save installation directory and use it in consequent install call.

`pip update` checks for new version and if it's available, update is simple wrapper around `pip remove` and `pip install`.

PR #272, opened 2017-09-11
[pip] Honor package installation directory during update

zrzka:
See #271. This PR gets package location just before `remove` is called and if location was found, sets `SITE_PACKAGES_FOLDER` before issuing `install` to keep updated package in the same folder. `pip install -d ...` works in the same way (also modifies `SITE_PACKAGES_FOLDER`). If location wasn't found, default one is used.

Tested ...

* installed package without `-d` -> `site-packages` -> `update` -> still `site-packages`
* installed package with `-d ~/Documents/site-packages-3` -> `update` -> still `site-packages-3`
* installed another package without `-d` -> `site-packages`

### DIFF
--- a/bin/pip.py
+++ b/bin/pip.py
@@ -852,6 +852,7 @@ class PackageRepository(object):
                                 and self.config.get_dependencies(dependency) is not None:
                             print('Removing dependency: {}'.format(dependency))
                             self.remove(dependency)
+
         else:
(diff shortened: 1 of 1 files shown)


fschaeck:
Changes from this pull request have been added via commit 9f8cb39 since this pull request couldn't be merge due to conflicts. The targeted issue #271 has been closed as well.

PR #286, opened 2018-02-10
Honor ~/.gitconfig for git user name, email

zed:
Don't ask for user name, email on the first git commit if they are already specified in the global user config: `~/.gitconfig`.


jsbain:
This fix only worked if .gitconfig already existed.  #333 fixes that.

PR #293, opened 2018-04-14
Start testing on Python 3 in allow_failures mode

cclauss:
It is time to make sure that StaSH can make the transition to Python 3.
* Also add flake8 tests in the before_script.


cclauss:
Closed in favor of #295

PR #295, opened 2018-04-14
Start testing on Python 3 in allow_failures mode

cclauss:
It is time to make sure that StaSH can make the transition to Python 3.
* Add flake8 tests in the Travis CI __before_script__.
* Resolve Python 2 undefined names (flake8 F821)
* Remove a few unused imports
* ~~My editor (Atom) automatically removes trailing whitespace~~
* Propose a solution to #296

Replaces both #293 and #294

@jsbain @zrzka @dgelessus Your reviews please

### DIFF
--- a/bin/crypt.py
+++ b/bin/crypt.py
@@ -13,10 +13,11 @@ optional arguments:
   -k KEY, --key KEY  Encrypt/Decrypt Key.
   -d, --decrypt      Flag to decrypt.
 '''
-import os
 import argparse
 import base64
+import os
 
+_stash = globals()['_stash']
 try:
 	import pyaes
 except:
--- a/bin/grep.py
+++ b/bin/grep.py
@@ -9,6 +9,7 @@ import re
 import sys
 
 def main(args):
+    global _stash
     ap = argparse.ArgumentParser()
     ap.add_argument('pattern', help='the pattern to match')
     ap.add_argument('files', nargs='*', help='files to be searched')
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
 language: python
 python:
   - "2.7"
+  - "3.6"
+matrix:
+  allow_failures:
+    - python: "3.6"
 # command to install dependencies
 install:
   - "pip install pyparsing==2.0.1"
-  - "pip install pytest>=2.8.1"
+  - "pip install pytest>=3.5.0"
+  - "pip install flake8>=3.5.0"
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 # command to run tests
 script: py.test system/tests --ignore=system/tests/data
--- a/bin/cp.py
+++ b/bin/cp.py
@@ -7,6 +7,8 @@ from __future__ import print_function
 import argparse
 import os
 import shutil
+import sys
+
 
 def pprint(path):
     if path.startswith(os.environ['HOME']):
--- a/bin/gh.py
+++ b/bin/gh.py
@@ -13,6 +13,10 @@ For all commands, use gh <command> --help for more detailed help
 NOTE: assumes a keychain user/pass stored in 	keychainservice='stash.git.github.com', which is also the default from the git module.  
 
 '''
+import os
+import sys
+
+
 def install_module_from_github(username, package_name, folder, version):
     """
     Install python module from github zip files
--- a/bin/git.py
+++ b/bin/git.py
@@ -23,16 +23,21 @@ Commands:
     help: git help
 '''
 
-SAVE_PASSWORDS = True
-
 import argparse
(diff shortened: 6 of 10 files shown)


bin/crypt.py
```diff
@@ -13,23 +13,24 @@
   -k KEY, --key KEY  Encrypt/Decrypt Key.
   -d, --decrypt      Flag to decrypt.
 '''
-import os
 import argparse
 import base64
+import os
 
+_stash = globals()['_stash']
```
dgelessus: I'm fairly certain this does nothing - `globals()` is always the current global namespace, which we are in right now, so this is equivalent to `_stash = _stash`.

bin/grep.py
```diff
@@ -9,6 +9,7 @@
 import sys
 
 def main(args):
+    global _stash
```
dgelessus: This is not needed - `_stash` is never assigned inside the function, so all uses of `_stash` already refer to the global name by default.

dgelessus:
OK, it seems that `flake8` (not `pyflakes`) provides an option to add extra global/builtin names. Adding `--builtins=_stash` should work.

cclauss:
@dgelessus Thanks for the great review.  All good and correct observations.  I will try your whitespace commit trick after I get a bit more caffeinated.  The other approach on whitespaces is to run __autopep8__ or __yapf__ across the whole codebase to fix whitespace and most of the other issues raised in the second flake8 run.  However that approach is often rejected by project maintainers for fear of “polluting” the commit history.

The other alternatives on undefined names are to put __#noqa: E821__ at the end of each offending line or to add a __.flake8__ config file to the root of the repo.  I usually take the _explicit is better that implicit_ approach of changing the code to be clear that the variable is defined in a global scope but I can be flexible.  I find that the __#noqa__ approach is usually rejected by those who have no prior experience with flake8 and it can be chatty / distracting for readers of the code.

Putting __\_\_builtins=_stash__ into the flake8 commands or .flake8 config file also works but it moves the assumptions out of the code and into a separate file which complicates moving code from one repo to another.  My preferred approach would be __from stash import _stash__ if that works because it is explicit yet no one would complain.

In general, flake8’s _undefined name_ capability is extremely valuable an I have found no other Python linter provides a similar check.  When porting to Python 3 undefined names can flag instances of __basestring, raw_input(), reload(), unicode, xrange()__, etc. that have all been moved or removed in Python 3.  In general usage, undefined names can find typos, scoping errors, refactoring errors, and missing imports that have the potential to raise __NameError__ at runtime.

bin/crypt.py
cclauss: There were existing instances of this elsewhere in the code so I replicated it because it makes the assumptions explicit to the reader.  Would __from stash import \__stash__ work as alternative?

bin/grep.py
cclauss: I actually like this approach and using _explicit is better than implicit_ as a justification, I am not inclined to change it unless __from stash import \_stash__ works.

cclauss:
I went back to the current master (__before__ this PR) and there are [4 pages of hits of using the __\_stash = globals()['\_stash']__ syntax](https://github.com/ywangd/stash/search?q=globals()&type=Code) so I am going to stick with copying that unless someone suggests a cleaner approach...

cclauss:
@dgelessus I finally got your whitespace restore trick to work...  I needed to __git rebase -i master__ and __drop__ the "remove trailing whitespace" commit.  Cool tip.  Thanks.

bin/crypt.py
ywangd: It indeed does nothing. I added it mainly to keep the IDE happy.

ywangd:
The trailing whitespace issue seems to be fixed? If so I’ll merge this PR.

bin/crypt.py
cclauss: ;-)  IDE probably runs flake8 or some other linter under the covers.

cclauss:
Yes.  Fixed.

dgelessus:
> My preferred approach would be *from stash import _stash* if that works because it is explicit yet no one would complain.

That sounds like a good idea. That way `_stash` is defined in the code instead of a config file, and the definition doesn't look like a no-op (like `_stash = globals()['_stash']` does).

bennr01:
Using 'from stash import _stash' may be problematic when running multiple stash instances in parallel.

cclauss:
> _stash is defined in the code instead of a config file
What config file?

> Using 'from stash import _stash' may be problematic when running multiple stash instances in parallel.
Running two StaSH instances in parallel should not be supported without a lot of testing.  Corner cases galore.

dgelessus:
> > _stash is defined in the code instead of a config file
>
> What config file?

Whereever you'd add the `--builtins=_stash` switch: `setup.cfg`, `.travis.yml`, `.flake8`, etc.

cclauss:
__get_stash()__ is https://github.com/ywangd/stash/blob/master/lib/stashutils/core.py#L16

bennr01:
@cclauss thanks :)
Would that be a suitable alternative to `from stash import _stash`?

ywangd:
@cclauss Could you make this pull request to the `dev` Instead of `master`? This is the normal workflow. Once changes are stable enough, they’ll be merged into master for releasing. Thanks.

cclauss:
Closed in favor of #299

PR #297, opened 2018-04-14
make stash py3 compatible

bennr01:
Hi,
this pull request attempts to make StaSh both py2 and py3 compatible.
It still contains some bugs and many commands are still incompatible with py3, however i dont think a single person can update stash to py3 alone (e.g. i am unable to port `git` or fix `ctype` threads).
I hope this PR can provide a solid base for completely making StaSh py3 compatible.

**Status:**
- stash core works in general (with some bugs) in both py2 and py3
- some base commands (`ls`, `cp`, ...) ported
- no *big* commands ported (`git`, `gh`, `mount`, ...)
- nearly full compatibility with py2

**Important information**:
- I hade to rename `stash.py` to `core.py` due to import issues, but if you want i can try to rename it back (also needs changes to `selfupdate.py`, ... )
- due to the change above, the old `selfupdate` can no longer check the version of the new stash version and an update needs to be forced using `selfupdate -f`.

**Known bugs**:
- `ctype` are bugged; they crash pythonista when killed
- `clear` does print a newline
- encoding of `sys.argv` for commands needs to be improved; currently there is only a quick fix in `shruntime.py`, which encodes sys.argv with `utf-8`. 

**pip changes:**
- actually print more informations (like traceback) if `--verbose` was specified.
- in py2, install packages to `site-packages-2`; in py3 to `site-packages-3` instead. Maybe we should add an argument to use  `site-packages` instead?

**Other changes:**
- add `python3` command
- add a `STASH_PY_VERSION` environment variable, which is also printed by `version.py`.
- minor changes

For more information, please see `$STASH_ROOT/docs/py3_todo.py`

### DIFF
--- a/bin/clear.py
+++ b/bin/clear.py
@@ -1,2 +1,4 @@
 """Clear the stash console output window"""
-print u'\u009bc',
+# TODO: prevent linebreak after 0099bc
+# This was removed as a quick fix
+print(u'\u009bc')
--- a/bin/crypt.py
+++ b/bin/crypt.py
@@ -20,7 +20,7 @@ import base64
 try:
 	import pyaes
 except:
-	print 'Installing Required packages.'
+	print('Installing Required packages...')
 	_stash('pip install pyaes')
 	import pyaes
 
@@ -69,4 +69,4 @@ if __name__ == '__main__':
 	else:
 		nk = crypt.aes_encrypt(args.key)
 		if args.key is None:
-			print "Key: ", nk
+			print("Key: %s" % nk)
--- a/bin/dropbox_setup.py
+++ b/bin/dropbox_setup.py
@@ -28,11 +28,11 @@ class DropboxSetupCmd(cmd.Cmd):
 
 	def do_list(self, cmd):
 		"""list: lists the dropbox usernames."""
-		print
+		self.stdout.write("\n")
 		for service, account in keychain.get_services():
 			if service == dbutils.DB_SERVICE:
-				print account
-		print
+				self.stdout.write(account + "\n")
+		self.stdout.write("\n")
 
 	def do_del(self, cmd):
 		"""del USERNAME: resets the dropbox for USERNAME."""
--- a/bin/edit.py
+++ b/bin/edit.py
@@ -14,6 +14,10 @@ import editor
 import time
 import argparse
 
+_stash = globals()["_stash"]
+
+if _stash.PY3:
+	raw_input = input
 
 def open_temp(file='', new_tab=True):
     try:
@@ -30,7 +34,7 @@ def open_temp(file='', new_tab=True):
             temp.flush()
             to_edit.close()
             
-        print '***When you are finished editing the file, you must come back to console to confim changes***'
+        print('***When you are finished editing the file, you must come back to console to confim changes***')
         editor.open_file(temp.name, new_tab)
         time.sleep(1.5)
         console.hide_output()
@@ -51,13 +55,13 @@ def open_temp(file='', new_tab=True):
                 with open(temp.name,'r') as tmp:
                     f.write(tmp.read())
                     
-            print 'File Saved.'
+            print('File Saved.')
         elif input=='N' or input=='n':
             if not new_tab:
                 editor.open_file(cur_path) # restore previous script in editor
     
-    except Exception, e:
-        print e
+    except Exception as e:
+        print(e)
         
     finally:
         temp.close()
--- a/bin/pip.py
+++ b/bin/pip.py
@@ -26,6 +26,7 @@ import contextlib
 import requests
 import re
 import operator
+import traceback
 
 import six
 from distutils.util import convert_path
@@ -51,11 +52,19 @@ PACKAGE_NAME_FIXER = {
     'lazy_object_proxy': 'lazy-object-proxy',
 }
 
-SITE_PACKAGES_FOLDER = os.path.expanduser('~/Documents/site-packages')
+_stash = globals()['_stash']
+
+if _stash.PY3:
+    unicode = str
+    str = bytes
+    SITE_PACKAGES_DIR_NAME = 'site-packages-3'
+else:
+    SITE_PACKAGES_DIR_NAME = 'site-packages-2'
+
+SITE_PACKAGES_FOLDER = os.path.expanduser('~/Documents/{}'.format(SITE_PACKAGES_DIR_NAME))
 
 NO_OVERWRITE = False
 
-_stash = globals()['_stash']
 
 
 def _setup_stub_(*args, **kwargs):
@@ -145,9 +154,9 @@ class PackageFinder(object):
     def _find_packages_iter(cls, base_path):
         candidates = cls._candidate_dirs(base_path)
         return (
-            path.replace(os.path.sep, '.')
-            for path in candidates
-            if cls._looks_like_package(os.path.join(base_path, path))
+        path.replace(os.path.sep, '.')
+        for path in candidates
+        if cls._looks_like_package(os.path.join(base_path, path))
         )
 
     @staticmethod
@@ -211,48 +220,48 @@ def fake_setuptools_modules():
     Created a bunch of stub setuptools modules
     """
     setuptools_modules = [
-        'setuptools',
-        'setuptools.command',
-        'setuptools.command.alias',
-        'setuptools.command.bdist_egg',
-        'setuptools.command.bdist_rpm',
-        'setuptools.command.bdist_wininst',
-        'setuptools.command.build_ext',
-        'setuptools.command.build_py',
-        'setuptools.command.develop',
-        'setuptools.command.easy_install',
-        'setuptools.command.egg_info',
-        'setuptools.command.install',
-        'setuptools.depends.install_egg_info',
-        'setuptools.command.install_lib',
-        'setuptools.command.install_scripts',
-        'setuptools.command.register',
-        'setuptools.command.rotate',
-        'setuptools.command.saveopts',
-        'setuptools.command.sdist',
-        'setuptools.command.setopt',
-        'setuptools.command.test',
-        'setuptools.command.upload',
-        'setuptools.command.upload_docs',
-        'setuptools.extern',
-        'setuptools.dist',
-        'setuptools.extension',
-        'setuptools.launch',
-        'setuptools.lib2to3_ex',
-        'setuptools.msvc9_support',
-        'setuptools.package_index',
-        'setuptools.py26compat',
-        'setuptools.py27compat',
-        'setuptools.py31compat',
-        'setuptools.sandbox',
-        'setuptools.site-patch',
-        'setuptools.ssl_support',
-        'setuptools.unicode_utils',
-        'setuptools.utils',
-        'setuptools.version',
-        'setuptools.windows_support',
-        # 'pkg_resources',
-        # 'pkg_resources.extern',
+    'setuptools',
+    'setuptools.command',
+    'setuptools.command.alias',
+    'setuptools.command.bdist_egg',
+    'setuptools.command.bdist_rpm',
+    'setuptools.command.bdist_wininst',
+    'setuptools.command.build_ext',
+    'setuptools.command.build_py',
+    'setuptools.command.develop',
+    'setuptools.command.easy_install',
+    'setuptools.command.egg_info',
+    'setuptools.command.install',
+    'setuptools.depends.install_egg_info',
+    'setuptools.command.install_lib',
+    'setuptools.command.install_scripts',
+    'setuptools.command.register',
+    'setuptools.command.rotate',
+    'setuptools.command.saveopts',
+    'setuptools.command.sdist',
+    'setuptools.command.setopt',
+    'setuptools.command.test',
+    'setuptools.command.upload',
+    'setuptools.command.upload_docs',
+    'setuptools.extern',
+    'setuptools.dist',
+    'setuptools.extension',
+    'setuptools.launch',
+    'setuptools.lib2to3_ex',
+    'setuptools.msvc9_support',
+    'setuptools.package_index',
+    'setuptools.py26compat',
+    'setuptools.py27compat',
+    'setuptools.py31compat',
+    'setuptools.sandbox',
+    'setuptools.site-patch',
+    'setuptools.ssl_support',
+    'setuptools.unicode_utils',
+    'setuptools.utils',
+    'setuptools.version',
+    'setuptools.windows_support',
+    # 'pkg_resources',
+    # 'pkg_resources.extern',
     ]
 
     for m in setuptools_modules:
@@ -263,21 +272,21 @@ def fake_setuptools_modules():
     import distutils.core
     import distutils.util
     distutils_command_modules = [
-        'distutils.command.bdist'
-        'distutils.command.bdist_dumb',
-        'distutils.command.bdist_msi',
-        'distutils.command.bdist_rpm',
-        'distutils.command.bdist_wininst',
-        'distutils.command.build',
-        'distutils.command.build_clib',
-        'distutils.command.build_ext',
-        'distutils.command.build_py',
-        'distutils.command.build_scripts',
+    'distutils.command.bdist'
+    'distutils.command.bdist_dumb',
+    'distutils.command.bdist_msi',
+    'distutils.command.bdist_rpm',
+    'distutils.command.bdist_wininst',
+    'distutils.command.build',
+    'distutils.command.build_clib',
+    'distutils.command.build_ext',
+    'distutils.command.build_py',
+    'distutils.command.build_scripts',
     ]
     for m in distutils_command_modules:
         fake_module(m)
     sys.modules['distutils.util'].get_platform = OmniClass()
-    # fix for new problem in issue 169 
+    # fix for new problem in issue 169
     sys.modules['distutils.command.build_ext'].sub_commands = []
     sys.modules['setuptools.command.build_ext'].sub_commands = []
 
@@ -343,9 +352,9 @@ class CIConfigParer(SafeConfigParser):
         section_name = self._get_section_name(name)
         return SafeConfigParser.items(self, section_name)
 
-    def get(self, name, option_name):
+    def get(self, name, option_name, *args, **kwargs):
         section_name = self._get_section_name(name)
-        return SafeConfigParser.get(self, section_name, option_name)
+        return SafeConfigParser.get(self, section_name, option_name, *args, **kwargs)
 
     def set(self, name, option_name, value):
         section_name = self._get_section_name(name)
@@ -361,10 +370,12 @@ class PackageConfigHandler(object):
     Manager class for packages files for tracking installation of modules
     """
 
-    def __init__(self):
-        self.package_cfg = os.path.expanduser('~/Documents/site-packages/.pypi_packages')
+    def __init__(self, verbose=False):
+        self.verbose = verbose
+        self.package_cfg = os.path.join(SITE_PACKAGES_FOLDER, '.pypi_packages')
         if not os.path.isfile(self.package_cfg):
-            print('Creating package file')
+            if self.verbose:
+                print('Creating package file...')
             with open(self.package_cfg, 'w') as outs:
                 outs.close()
         self.parser = CIConfigParer()
@@ -440,8 +451,8 @@ class ArchiveFileInstaller(object):
     class SetupTransformer(ast.NodeTransformer):
         """
         Analyze and Transform AST of a setup file.
-            1. Create empty modules for any setuptools imports (if it is not already covered)
-            2. replace setup calls with a stub one to get values of its arguments
+        1. Create empty modules for any setuptools imports (if it is not already covered)
+        2. replace setup calls with a stub one to get values of its arguments
         """
 
         def visit_Import(self, node):
@@ -458,10 +469,10 @@ class ArchiveFileInstaller(object):
         def visit_Call(self, node):
             func_name = self._get_possible_setup_name(node)
             if func_name is not None and \
-                    (func_name == 'setup' or func_name.endswith('.setup')):
+            (func_name == 'setup' or func_name.endswith('.setup')):
                 node.func = ast.copy_location(
-                    ast.Name('_setup_stub_', ast.Load()),
-                    node.func)
+                ast.Name('_setup_stub_', ast.Load()),
+                node.func)
             return node
 
         def _get_possible_setup_name(self, node):
@@ -477,6 +488,10 @@ class ArchiveFileInstaller(object):
             else:
                 return None
 
+
+    def __init__(self, verbose=False):
+        self.verbose = verbose
+
     def run(self, pkg_name, archive_filename):
         """
         Main method for Installer to do its job.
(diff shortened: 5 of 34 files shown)


cclauss:
Wow!  Nice work.  Once #295 landed, I was going to propose [a more automated approach](https://github.com/cclauss/itinerant-futurizer) but let's see how this PR goes.  Please add [.travis.yml changes](https://raw.githubusercontent.com/cclauss/stash/dc0483de9b3be1b82bc94329462a4e311bfe39ec/.travis.yml) so we can see how things are progressing.

bennr01:
@cclauss thanks. #295 reminded me that i still have to do this PR :smile:
I will try to add the changes to `.travis.yml`, but it seems that this PR actually broke some the old tests anyway.

bin/edit.py
```diff
@@ -14,6 +14,10 @@
 import time
 import argparse
 
+_stash = globals()["_stash"]
+
+if _stash.PY3:
```
cclauss: @ywangd and @bennr01 Do you have any objections to us using the very popular [__six__ module](http://six.readthedocs.io) to achieve compatibility with both Python 2 and Python 3?  It will be much cleaner than a bunch of __if PY3__ statements and complex workarounds for things that are in __six.moves__.

bennr01: @cclauss no objection by me. I only decided against using `six` as i have no experience with it.

bin/clear.py
```diff
@@ -1,2 +1,4 @@
 """Clear the stash console output window"""
-print u'\u009bc',
+# TODO: prevent linebreak after 0099bc
+# This was removed as a quick fix
+print(u'\u009bc')
```
dgelessus: `print(..., end="")` (requires `from __future__ import print_function`)

bin/dropbox_setup.py
```diff
@@ -28,11 +28,11 @@ def do_exit(self, cmd):
 
 	def do_list(self, cmd):
 		"""list: lists the dropbox usernames."""
-		print
+		self.stdout.write("\n")
```
dgelessus: `print("")`

(or simply `print()` with `from __future__ import print_function`)

bin/edit.py
dgelessus:
```python
try:
    raw_input
except NameError:
    raw_input = input
```

No `six` or `PY3` required :)

bin/find.py
```diff
@@ -112,7 +112,7 @@ def main(args):
     ap.add_argument('-maxdepth', '--maxdepth',
                     metavar='n',
                     nargs='?',
-                    default=sys.maxint,
+                    default=(2 ** 30),
```
dgelessus: `sys.maxsize` (means a slightly different thing than Python 2's `sys.maxint`, but works well as an "almost infinity" number)

Alternatively, default to `None`, and skip the depth check in that case.

bin/gci.py
```diff
@@ -1,104 +1,106 @@
...
-			for k in dir(gc.garbage[0]):
-				try:
-					delattr(gc.garbage[0], k)
-				except:
-					pass
-			del gc.garbage[:]
-
-if __name__ == "__main__":
-	main()
+# coding: utf-8
```
dgelessus: GitHub's diff viewer shows that every line in this file has changed, but most of them look the same. Any idea what happened here? (Did the end-of-line type change perhaps?)

bin/pip.py
```diff
@@ -51,11 +52,19 @@
     'lazy_object_proxy': 'lazy-object-proxy',
 }
 
-SITE_PACKAGES_FOLDER = os.path.expanduser('~/Documents/site-packages')
+_stash = globals()['_stash']
+
+if _stash.PY3:
+    unicode = str
+    str = bytes
```
dgelessus: `str = bytes` doesn't seem like a good idea, that will be very confusing for people who expect Python 3 code (where `str` means `unicode`). I think it would be better to always use `unicode` and `bytes` explicitly. That way it's always clear which type of string is meant.

Actually, there is a case where you want to use `str` instead of `bytes`/`unicode`: some strings have to have the "default string" type (bytestring in Python 2 and Unicode string in Python 3), for example the attribute name passed to `getattr`. This is another reason why overwriting `str` would be problematic, because then there's no easy way to use the "default string" type.

bin/wget.py
```diff
@@ -24,17 +29,20 @@ def main(args):
     console.show_activity()
     try:
 
-        print 'Opening: %s' % url
+        sys.stdout.write('Opening: %s\n' % url)
```
dgelessus: I think all of these `sys.stdout.write` calls could be translated to `print` calls.

core.py
```diff
@@ -9,8 +9,19 @@
...
+
+try:
+	# try to import py2 modules
+	from ConfigParser import ConfigParser
+	from StringIO import StringIO
+	PY3 = False
+except ImportError:
+	# py 3
+	from configparser import ConfigParser
+	from io import StringIO, IOBase as file
```
dgelessus: The `io` module exists under Python 2 as well and is basically identical. It might be better to always use the types from `io` instead of the legacy Python 2 names.

system/shruntime.py
```diff
@@ -476,7 +487,13 @@ def exec_py_file(self, filename,
 
         saved_sys_argv = sys.argv[:]
         # First argument is the script name
-        sys.argv = [os.path.basename(filename)] + (args or [])
+        argv = [os.path.basename(filename)] + (args or [])
+        
+        # convert sys.argv to unicode if on python3
+        if PY3:
+        	# todo: this is only a temporary solution and needs to be redone. This may be buggy.
+        	argv = [c if isinstance(c, unicode) else c.decode("latin-1") for c in argv]
```
dgelessus: `"latin-1"` is not a good default encoding, it only works for a very small number of characters and is almost never used anymore. `"utf-8"` is the de facto standard, that would be a better choice. (Or simply leave out the encoding completely - Python defaults to `"utf-8"` if you don't pass an encoding to `decode`/`encode`.)

bin/edit.py
dgelessus: To clarify - I don't have anything against `six`, but personally I've found it overkill in most cases. The most common missing names are `unicode`, `raw_input`, and `xrange`, which can be defined when needed with a `try`-`except NameError` block. Renamed modules can be handled with `try`-`except ImportError` as well. (`urllib` and friends are perhaps the most annoying case, in that case `six` is useful.)

Also I think `six` is more necessary when you need to target ancient Python versions (older than 2.7 or 3.3). These are missing various useful 2/3 compatibility features, which `six` can emulate. 2.7 is very forward-compatible, so there aren't many things that `six` can do that you can't do yourself in a couple of lines.

Regardless of whether or not you're using `six`, `__future__` imports are very useful too. `from __future__ import absolute_import, division, print_function` is a good default for Python 2/3 code, it makes some things behave more like Python 3 even in Python 2.

cclauss: Great observations...  I tend to lean a bit more towards six than you do because it does not slow down the reader of the code but your alternatives are all correct.

bin/pip.py
cclauss: These are the kind of places where six shines...

system/shruntime.py
cclauss: +1 on utf-8

bin/clear.py
bennr01: Huh, didnt know that you could use `__future__` imports from withing StaSh. Seems like there is still so much to learn :)

bin/dropbox_setup.py
bennr01: I tend to use `self.stdout` instead of `print` in `cmd.Cmd` subclasses for code reusability. `self.stdout` is an attribute of `cmd.Cmd` and may not be `sys.stderr`.

bin/gci.py
bennr01: I have no idea. Maybe some tab/space replacement?
I think that also happend in some other file.

system/shruntime.py
bennr01: IIRC `"latin-1"` has better bytes compatibility. Decoding a random bytestring (e.g. `os.urandom(128)`) in `"utf-8"` will likely raise an Error, but `"latin-1"` can handle it.
Anyway, this was only meant as a quick fix and the encoding should probably be moved to to the `stash.system.*` module responsible for the input.

core.py
```diff
@@ -277,7 +277,7 @@ def text_style(s, style, always=False):
         # No color for pipes, files and Pythonista console
         if not always and (
         	isinstance(sys.stdout, StringIO)
-        	or isinstance(sys.stdout, file)
+        	or isinstance(sys.stdout, IOBase)
```
cclauss: The second parameter of [__isinstance()__](https://docs.python.org/2/library/functions.html#isinstance) can be a tuple as in __isinstance(sys.stdout, (StringIO, IOBase))__.

However you do not even need that trick here because StringIO is inherited from IOBase...
```
$ python3
>>> import io
>>> sio = io.StringIO()
>>> isinstance(sio, io.IOBase)
True
```

ywangd:
Thanks for all the pull requests. Python 3 compatibility is indeed important. 

I see there are some Travis CI errors for this pull request. Should they be fixed before I merge it?

cclauss:
Yes.  I think we should keep Travis happy before merging.

bennr01:
@ywangd @cclauss It seems most of the current errors have been fixed by @cclauss in #295. I think fixing them also in this PR may cause a merge conflict, but i am not sure.
Also, when looking at [the Travis CI build for an earlier commit](https://travis-ci.org/ywangd/stash/builds/366457102?utm_source=github_status&utm_medium=notification) there are a whole lot of other errors in the tests and for some of them i am not sure how to fix them (e.g. the errors related to stdout being mixed up; the only real change done to the stdout mechanics where replacing `file` with `io.IOBase` in `system/shruntime.py`, but that is required).
I will take a close look and see what i can do.

cclauss:
Let's wait for @ywangd to merge #295 and then rebase on master and see how the tests come out.

bin/dropbox_setup.py
dgelessus: Ah, I misread that - I thought it said `sys.stdout`. I didn't know that `self.stdout` existed here. Thanks for the explanation!

bin/pip.py
```diff
@@ -684,6 +705,8 @@ def _run_setup_file(filename):
                 print("Warning: pkg_resources not available, skipping entry_points definitions.")
                 entry_points = {}
         for epn in entry_points:
+            if self.verbose:
+                print("Handling entrypoints for: " + epn)
             ep = entry_points[epn]
             if isinstance(ep, (str, unicode)):
```
cclauss: if isinstance(ep, basestring):

system/shruntime.py
```diff
@@ -20,10 +25,14 @@
...
 from .shcommon import _STASH_ROOT, _STASH_HISTORY_FILE, _SYS_STDOUT, _SYS_STDERR
-from .shcommon import is_binary_file, _STASH_EXTENSION_BIN_PATH
+from .shcommon import is_binary_file, _STASH_EXTENSION_BIN_PATH, PY3
 from .shparsers import ShPipeSequence
 from .shthreads import ShBaseThread, ShTracedThread, ShCtypesThread, ShState, ShWorkerRegistry
 
+if PY3:
+	# rename unicode, str
+	unicode = str
+	str = bytes
```
cclauss:
```python
try:
    unicode        # Python 2
except NameError:
    unicode = str  # Python 3
```

system/shterminal.py
```diff
@@ -12,7 +12,13 @@
...
     from .dummyobjc_util import *
 
-from .shcommon import CTRL_KEY_FLAG
+from .shcommon import CTRL_KEY_FLAG, PY3
+
+
+if PY3:
+	# rename string types
+	unicode = str
+	str = bytes
```
cclauss:
```python
try:
    unicode        # Python 2
except NameError:
    unicode = str  # Python 3
```

lib/libcore.py
```diff
@@ -1,6 +1,13 @@
 import os
 import fileinput
 
+from stash.system.shcommon import PY3
+
+
+if PY3:
+	# py3 compatibility
+	unicode = str
+
```
cclauss:
```python
try:
    unicode        # Python 2
except NameError:
    unicode = str  # Python 3
```

bin/pip.py
cclauss:
```python
try:
    unicode        # Python 2
except NameError:
    unicode = str  # Python 3
```

system/shscreens.py
```diff
@@ -15,10 +15,15 @@
...
 
-from .shcommon import IN_PYTHONISTA, ON_IOS_8
+from .shcommon import IN_PYTHONISTA, ON_IOS_8, PY3
 # noinspection PyPep8Naming
 from .shcommon import sh_delay, Graphics as graphics
 
+
+if PY3:
+	# set xrange as alias for range
+	xrange = range
```
cclauss:
```python
try:
    xrange          # Python 2
except NameError:
    xrange = range  # Python 3
```

system/shterminal.py
bennr01: Ok, i get your point about `try: name except NameError: name = ...`.
I just dislike using `try: ... Except` if i already know that i have to use python3 ;)
Though i probably should do these changes since you likely have way more experience with python3 than me.

core.py
```diff
@@ -88,6 +101,8 @@ class StaSh(object):
     Main application class. It initialize and wires the components and provide
     utility interfaces to running scripts.
     """
+    
+    PY3 = PY3
```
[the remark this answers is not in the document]
bennr01: The `PY3` on the left side is a attribute of the `StaSh()` class and its instances, the right side is a module-level `PY3` constants.
Both have a reason to exist: Because we need to check for `PY3` before creating the `StaSh` class, the module level one exists. And because `if _stash.PY3:` is a simpler syntax than importing or defining a custom `PY3` variable in each command that one also exists.

core.py
```diff
@@ -69,7 +80,9 @@
 BUFFER_MAX=150
 AUTO_COMPLETION_MAX=50
 VK_SYMBOLS=~/.-*|>$'=!&_"\?`
-""".format(text_size=14 if ON_IPAD else 12)
+""".format(
+	font_size=(14 if ON_IPAD else 12),
+	)
```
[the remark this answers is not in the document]
bennr01: This change is a relict from earlier rewrite attempts. I made the first commit for py3 compatibility once the stash core war more or less working, so i had to manually undo any unneccessary changes.
IIRC, at some point i tried using explicit bytestrings for the default config. Apparently python 3 doesn't like `.format()` on bytestrings, so i had to switch over to `%s` formating (which - for some weird reasons - works on py3 bytestrings even though it was at some point deprecated).
In the end i tried to undo the changes without having access to the old source and then this happened.

bin/wget.py
```diff
@@ -49,11 +58,12 @@ def main(args):
...
-                print '\r' + status,
-            print
+                print('\r' + status, end="")
+            print("")
 
-    except:
-        print 'invalid url: %s' % url
+    except Exception as e:
+        raise e
+        print('invalid url: %s' % url)
```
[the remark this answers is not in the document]
bennr01: Oops, seems like i forgot to remove that line ;)

system/shterminal.py
cclauss: Python porting best practice [use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection)

ywangd:
@dgelessus For checking whether output is a file, one reason that I can think of is to disable colours. Colour encoding is only active when output is terminal.

bin/pip.py
dgelessus: `basestring` doesn't exist in Python 3.

cclauss: [__six.string_types__](http://six.readthedocs.io/#six.string_types) ;-)

dgelessus:
That makes sense. Should we perhaps add a utility function somewhere that returns whether a file is the Stash terminal? That would be easier to understand than an `isinstance` check, and it would be easier to change in the future if the `isinstance` check doesn't work anymore for some reason.

bin/selfupdate.py
```diff
@@ -21,7 +21,8 @@
 
 
 class UpdateError(Exception):
-    pass
+    def __init__(self, message=""):
+    	self.message = message
```
dgelessus: Wouldn't it be better to change the code that relies on the `message` attribute instead? If you want to display an exception, it's best to use `str(exc)`, so the name of the exception is included. If you want to leave the exception name out, you can use `exc.args[0]` (which works in Python 2 and 3).

bennr01: @dgelessus thanks for the tip :)

system/shterminal.py
bennr01: @cclauss thanks for the link. This is my first time porting something to python3 so i am a bit inexperienced ;)

system/shparsers.py
```diff
@@ -5,7 +5,11 @@
 import glob
 import logging
 import threading
-from StringIO import StringIO
+try:
+	from StringIO import StringIO
+except ImportError:
+	# py3
+	from io import StringIO
```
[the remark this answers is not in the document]
bennr01: @cclauss unfortunately, that wont work here. We need a `StringIO` for native strings (bytestrings in py2 and unicode strings in py3).

core.py
```diff
@@ -249,9 +275,12 @@ def text_style(s, style, always=False):
         :return:
         """
         # No color for pipes, files and Pythonista console
-        if not always and (isinstance(sys.stdout, StringIO)
-                           or isinstance(sys.stdout, file)
-                           or sys.stdout.write.im_self is _SYS_STDOUT):
```
bennr01: Just remembered: has anyone any idea what line 254 in `stash.py` (now `core.py`) did?
Default `sys.stdout.write` does not seem to have an `im_self` attribute and it is nowhere defined in the StaSh code, so i assumed it was a buggy line which just did not execute due to `or`-magic and commented it out.

ywangd: This is to check whether the output is Pythonista console. If it is, the value will be a StdoutCatcher object, which is only available in Pythonista.

ywangd:
@bennr01 Just merged @cclauss PR, could you please check again whether the errors are fixed?

core.py
bennr01: @ywangd thanks, i didnt check that. Should i add it back?

ywangd: Yes please.

bennr01:
@ywangd Py2 tests are now working and complete successful.
Python3 tests still fail but travis is happy anyway.
Should i try to fix them before the merge?
Also, i still have to add the 'im_self' check back.

cclauss:
If you want to reduce the flake8 error clutter, you can a fix to only the __print()__ issues with:
* __pip install future__
* __futurize -f libfuturize.fixes.fix_print_with_import -w .__

bin/selfupdate.py
cclauss: __HINT:__ Undo these modifications and then in __atom system/shruntime.py__ change all instances of __e.message__ to __e.args[0]__

system/shparsers.py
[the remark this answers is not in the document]
cclauss: http://six.readthedocs.io/#six.StringIO

bin/crypt.py
```diff
@@ -19,12 +19,11 @@
...
 _stash = globals()['_stash']
 try:
-    import pyaes
-except Exception:
-    print 'Installing Required packages.'
-    _stash('pip install pyaes')
-    import pyaes
-
+	import pyaes
+except:
```
cclauss: except Exception:

bin/tar.py
```diff
@@ -93,7 +93,7 @@ def tar_filter(tarinfo):
         output_print('Adding %s' % name)
         tar.add(name, filter=tar_filter)
     tar.close()
-    print 'Archive Created.' 
+    print('Archive Created.' )
```
cclauss: Trailing whitespace inside parens  ;-)

docs/py3_todo.md
```diff
@@ -0,0 +1,29 @@
+#Python 3 TODO
+List of known issues, bugs and todos for stash py3 (and py2) compatibility.
+
+#General
+- ctype threads in py3 bugged (crash on cc and kill)
+- more commands need to be ported
+- sys.argv needs to be bytestr in py2 and unistt in py3. There is only a quick fix in place (somewhere in `shruntime.py`), which should be replaced as it only works for common usage situations
```
cclauss: Would this do the trick?  __args = [six.text_type(arg) for arg in sys.argv]__?

docs/py3_todo.md
```diff
@@ -0,0 +1,29 @@
...
+- more testing
+
+#`easy_config.py`
+- while syntax is py3 compatible, there are some errors
+- untested
+
+#`pip.py`
+- many bugs
+- installation using `setupy.py` fails most of the time, always falls back to package detection
+- maybe `pip3` for py3 instead of a single `pip` command.
```
cclauss: If we _need_ two different pips, then we should make three instead (like homebrew does)?
* __pip2__ always installs into site packages for Python 2
* __pip3__ always installs into site packages for Python 3
* __pip__ looks to see what Python is currently running and then uses __pip2__ or __pip3__ accordingly.

system/shruntime.py
```diff
@@ -1,10 +1,15 @@
...
+import platform
 import logging
 import threading
 import functools
-from StringIO import StringIO
+
+try:
+	from StringIO import StringIO
+except ImportError:
+	from io import StringIO, IOBase as file
```
cclauss:
```python
from six import StringIO, PY3
try:
    file           # Python 2
except NameError:  # Python 3
    from io import IOBase as file
```

system/shscreens.py
```diff
@@ -19,6 +19,12 @@
 # noinspection PyPep8Naming
 from .shcommon import sh_delay, Graphics as graphics
 
+
+try:
+	xrange
+except NameError:
+	xrange = range
```
cclauss: __from six.moves import xrange__

system/tests/test_completer.py
```diff
@@ -53,7 +53,7 @@ def test_completion_08(self):
         assert newline == 'git branch '
 
     def test_completion_09(self):
-        newline, possibilities = self.complete('$STASH_')
+        newline, possibilities = self.complete('$STASH_R')
```
cclauss: Is there a __$STASH_R__defined or was this supposed to be __$STASH_ROOT__?

system/tests/test_expander.py
```diff
@@ -36,7 +36,7 @@ def test_wildcards(self):
         assert pipe_sequence.lst[0].args[0] == 'README.md'
 
         pipe_sequence = self._get_pipe_sequence(r'ls *stash*')
-        assert 'stash.py' in pipe_sequence.lst[0].args
+        # assert 'stash.py' in pipe_sequence.lst[0].args
```
cclauss: Why remove this assertion?

docs/py3_todo.md
bennr01: @cclauss `six.text_type` is `unicode` in py2, but we need a bytestring.

docs/py3_todo.md
bennr01: @cclauss Good idea. I dont think we actually *need* two different pips. It seems like we *could* just use the same one. The `pip3` proposal was only meant for consistency with the `python` and `python3` commands.

lib/stashutils/fsi/zip.py
```diff
@@ -6,7 +6,8 @@
 import shutil
 import datetime
 import stat
-import StringIO
+
+from io import BytesIO
```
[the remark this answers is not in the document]
bennr01: @cclauss is there any reason to use `six.BytesIO` instead of `io.BytesIO`? It seems like `io.BytesIO` is the same in both py2 and py3.

system/tests/test_completer.py
bennr01: @cclauss We test wether `$STASH_R` will be completed to `$STASH_ROOT`. This PR introduced an environment variable called `$STASH_PY_VERSION`, so the old `$STASH_` completion test completed to `$STASH_PY_VERSION` instead of `$STASH_ROOT` and failed.

system/tests/test_expander.py
bennr01: This test attempts to expand `*stash*` to `stash.py`, `getstash.py` and `launch_stash.py`.
`stash.py` was renamed to `core.py` to fix issues with the `import` mechanic changes in py3, so `*stash*` can no longer be completed to `stash.py`.
Since there is no reason to replace this assertion in this test, i removed it.

docs/py3_todo.md
cclauss: http://six.readthedocs.io/#six.binary_type ?

docs/py3_todo.md
cclauss: Good if we only have one pip command.  The Python command could operate the same way...
* __python2__ works on Python __2__ but fails with an error message on Python __3__
* __python3__ works on Python __3__ but fails with an error message on Python __2__
* __python__ looks to see what Python is currently running and then uses python2 or python3 accordingly.

Much like homebrew https://docs.brew.sh/Homebrew-and-Python

lib/stashutils/fsi/zip.py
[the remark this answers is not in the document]
cclauss: No problems.  Leave it as is.

system/tests/test_expander.py
cclauss: Would it be appropriate to have __assert 'core.py' in pipe_sequence.lst[0].args__

bennr01: @cclauss no. We test for expansion of `*stash*`, which can never be expanded to `core.py`.
An alternative would be adding a new test which checks for completion of `core*`, but since we dont test for the existence of that file but for completion, there is no reason to do so.

docs/py3_todo.md
dgelessus: By the way, this is specified by [PEP 394](https://www.python.org/dev/peps/pep-0394/); it's not specific to Homebrew.

docs/py3_todo.md
dgelessus: How about `str`?

ywangd:
@bennr01 It's ok to have some PY3 failures for the time being. The Travis CI change is to allow this anyway. Thanks for fixing all tests for PY2.

PR #307, opened 2018-04-18
Fix py3 tests; make most commands syntax compatible

bennr01:
Thanks to @cclauss, the py3 tests now pass successfully.

### DIFF
--- a/system/shcommon.py
+++ b/system/shcommon.py
@@ -10,6 +10,9 @@ import threading
 import ctypes
 from itertools import chain
 
+import six
+
+
 IN_PYTHONISTA = sys.executable.find('Pythonista') >= 0
 
 if IN_PYTHONISTA:
@@ -74,7 +77,7 @@ _EXTERNAL_DIRS = [
 	]
 
 # Python 3 or not Python 3
-PY3 = (sys.version_info >= (3, 0))
+PY3 = six.PY3
 
 # Save the true IOs
 if IN_PYTHONISTA:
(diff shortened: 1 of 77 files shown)


system/shruntime.py
```diff
@@ -576,6 +574,19 @@ def exec_sh_file(self, filename,
         except:
             self.stash.write_message('%s: error while executing shell script\n' % filename)
             current_state.return_value = 2
+    
+    def encode_argv(self, argv):
+    	"""
+    	Convert an argv list into the appropiate string type depending
+    	on the currently used python version.
+    	"""
+    	if PY3:
```
cclauss:
```python
argv_type = text_type if PY3 else binary_type
return [c if isinstance(c, argv_type) else c.encode("utf-8") for c in argv]
```

cclauss:
You can whack the whole matrix.

system/shcommon.py
```diff
@@ -177,7 +175,10 @@ def is_binary_file(filename, nbytes=1024):
     """
     with open(filename, 'rb') as ins:
         for c in ins.read(nbytes):
-            oc = ord(c)
+            if isinstance(c, six.integer_types):
+                oc = c
+            else:
+                oc = ord(c)
```
cclauss: oc = c in isinstance(c, int) else ord(c)  # does that not work?

bennr01: @cclauss Sure, it would work, but is there any reason for using compact syntax here? In most cases, it is a bit easier to read full `if`/`for`/`...` statements than to read their compact equivalents.

cclauss:
I like tighter code but it is personal preference.  __int__ will be faster than __six.integer_types__

PR #311, opened 2018-05-04
pip: fix https error in pip (fix 309)

### LINKED ISSUE #309: Error: Failed to fetch package release urls
i can't use pip install , it always show 'Error: Failed to fetch package release urls'  but i can use pip search , what happened?

bennr01:
This is a fix for the `Error: Failed to fetch package release urls` error in `pip`.
The bug was reported by @jphoebe4086824 and the fix was discovered by @nasmatic.
This PR fixes #309.

### DIFF
--- a/bin/pip.py
+++ b/bin/pip.py
@@ -924,7 +924,7 @@ class PyPIRepository(PackageRepository):
     def get_standard_package_name(self, pkg_name):
         if pkg_name not in self.standard_package_names:
             try:
-                r = requests.get('http://pypi.python.org/pypi/{}/json'.format(pkg_name))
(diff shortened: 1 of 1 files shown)


ywangd:
Thanks @bennr01 
Also do you think the dev branch is stable enough to merge into master for a Python 3 compatible release?

bennr01:
@ywangd common commands in dev should be pretty stable by now, but there may still be a few bugs., especially in `git` and `pip`. StaSh in py2 should still be stable. Should we just add a warning on startup that StaSh in py3 may be bugged?

ywangd:
Sure. Could you please add it? Also, could you please bump the version to 0.7.0? I think the py3 upgrade worth a new major version.

PR #312, opened 2018-05-04
add py3 warning on startup; version bump; minor bugfixes and improvements

bennr01:
Hi,
this PR adds a warning which will be shown if StaSh is started with python3.
It also contains the version bump to `0.7.0` (As discussed in #311 ).
Other than that, there are also some minor improvements and bugfixes regarding py3 compatibility.
Since the version detection of stash changed in a backwards incompatible way, a forceupdate (`selfupdate -f`) is required to update from  `0.6.20` or older to `0.7.0`)

Also, there seem to be a lot of issues which could be closed. Should i compile a list of all issues which can be closed?

### DIFF
--- a/bin/easy_config.py
+++ b/bin/easy_config.py
@@ -261,7 +261,7 @@ class ColorPicker(object):
 	@property
 	def hexcode(self):
 		"""returns the selected color as a html-like hexcode"""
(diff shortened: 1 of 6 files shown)


ywangd:
Thanks! Yes please close any issues that are no longer valid after 0.7.0

PR #332, opened 2018-08-27
Correct urlparse break from 3f958a9

jsbain:
Commit #3f958a9 replaced an import of urlparse module, urlparse function.  Not surprisingly, this breaks most uses of git.  This commit replaces urlparse usage throughout git.py.  This should be pulled into master asap.


TheLastDestroyer:
There’s also urlunparse which needs to be imported and changed from urlparse.urlunparse

jsbain:
Ok, should be good to go now

PR #347, opened 2018-12-30
Fix broken Text- & Backgroundcolor

bennr01:
After two years, I finally found the bug!
As mentioned in PR #207, the backgroundcolor (and textcolor) system was broken. When using non-black background-colors, the background would have the correct color, but the text would always have black background and white text.
This PR fixes this bug, but reduces the color range for the terminal from full RGB to the colors specified in `$STASH_ROOT/system/shscreens.py` (excluding `smoke`). Additionally, the values of some colors have been adjusted to match the terminal backgroundcolors.

**Fixes:**
- fix background and text colors
- fix tests for pip search (but will most likely break again in the future)

**Improvements:**
- `easy_config.py` was updated to reflect these changes
- `easy_config.py` will now report `$SELFUPDATE_TARGET`.
- The `.stash_config` files now accept all values which would be accepted by the `ui`-module. However, these would only work on the actual background and not on the text, so you should use strings instead (e.g. `"red"`).

### DIFF
--- a/bin/easy_config.py
+++ b/bin/easy_config.py
@@ -8,6 +8,7 @@ from six import string_types
 import console
 import pythonista_add_action as paa
 import ui
+import dialogs
 from stash.system.shcommon import _STASH_CONFIG_FILES
 
 _stash = globals()["_stash"]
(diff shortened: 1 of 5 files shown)


ywangd:
@bennr01 Is the failed test just being flaky?

ywangd:
@bennr01 Thanks I'll merge it then.

PR #352, opened 2019-01-22
Support multiple -H commands in curl.py to build composite request

whorfin:
Right now curl.py is not compatible with curl when there are multiple -H commands, as only the last one is used.  This fixes that, so -H commands build a composite set of headers, as is the typical use case.


whorfin:
The travis failure here seems to be due to a momentarily busted testing harness?

bennr01:
> The travis failure here seems to be due to a momentarily busted testing harness?

Yes, the failing test was fixed in https://github.com/ywangd/stash/commit/36c1c89c3760e7cbf013d1cb415c83a925543291, but your fork does not yet contain the fix. You can safely ignore the failing test.

Issue #364, opened 2019-03-23
BUG: class 'UnicodeDecodeError'(stash 0.7.2)


yjqiang:
```
[~/Documents]$ pip install requests
Querying PyPI ... 
Downloading package ...
Opening: https://files.pythonhosted.org/packages/7d/e3/20f3d364d6c8e5d2353c72a67778eb189176f08e873c9900e10c0287b84b/requests-2.21.0-py2.py3-none-any.whl

Save as: /private/var/mobile/Containers/Data/Application/106690BE-96ED-4657-8531-E99119A7CD8F/tmp//requests-2.21.0-py2.py3-none-any.whl (57987 bytes)
     57987  [100.00%]
Installing wheel: requests-2.21.0-py2.py3-none-any.whl...
<class 'UnicodeDecodeError'>: 'ascii' codec can't decode byte 0xe2 in position 1602: ordinal not in range(128)
[~/Documents]$ 

```
I'm using stash 0.7.2 with pythonista 3.2(320000)

yjqiang:
https://github.com/ywangd/stash/blob/master/lib/stashutils/wheels.py#L309
should be `with open(p, "r", encoding='utf-8') as fin:`

yjqiang:
closed

PR #367, opened 2019-03-26
Fix 'UnicodeDecodeError' when using pip install

yjqiang:
https://github.com/ywangd/stash/issues/364

### DIFF
--- a/lib/stashutils/wheels.py
+++ b/lib/stashutils/wheels.py
@@ -6,6 +6,7 @@ import json
 import re
 import zipfile
 import platform
+from io import open
 
 import six
 from six.moves import configparser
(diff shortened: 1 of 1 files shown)


yjqiang:
In python3 `open` from `io` is an alias for the built-in `open`. See docs.python.org/3/library/io.html?highlight=io#io.open – mfussenegger Jul 6 '16 at 20:27
https://stackoverflow.com/questions/10971033/backporting-python-3-openencoding-utf-8-to-python-2

bennr01:
Hi, thanks for the PR :) It seems to work, so i just merged it. I hope this is OK, @ywangd? Not sure if i am allowed to merge...

ywangd:
Thanks @bennr01 Please merge as you see fit.

Issue #372, opened 2019-05-17
Question: How to install without top_level.txt? (eg: Wheel generated by: flit)


yjqiang:
https://github.com/ywangd/stash/issues/370

yjqiang:
Sorry I don't know the details in pypi. So maybe we can just list all the file in the *.whl except the dist-info file and just move all files to site-packages?

bennr01:
This could be a solution. But we should not ignore `top_level.txt` if it does exists, so maybe use this as a fallback if the file does not exists.
This change would affect `$STASH_ROOT/lib/stashutils/wheels.py`. There is a class called `TopLevelHandler`, if you want to make the change.
I will look a bit more into the wheel format to check of there is a more specific rule for this case.

yjqiang:
https://paste.ubuntu.com/p/SHpzz32qTp/   I changed 174-185

PR #373, opened 2019-05-17
Fix wheels: Install without top_level.txt

yjqiang:
https://github.com/ywangd/stash/issues/372


bennr01:
Thanks for the fix. The changes seem good and the tests are working, so i am going to merge this.

PR #377, opened 2019-06-01
Add "pip uninstall" to libcompleter

yjqiang:


yjqiang:
PS: Looks like https://travis-ci.org/ywangd/stash/jobs/540022072#L9881 is not right, but I don't think it is my fault. 😂

bennr01:
@yjqiang thanks for the PR.  Seems like i have forgotten to change this when i renamed `pip remove` to `pip uninstall`. Thanks :smile:

> PS: Looks like https://travis-ci.org/ywangd/stash/jobs/540022072#L9881 is not right, but I don't think it is my fault. 😂

Yeah, `tests/system/test_threads.py::ThreadsTests::test_101` fails randomly. This happens often. We should probably mark it as an `xfail`.

By the way, could you please make future PRs to the `dev` branch? It does not realy matter, but this way we can check larger changes for bugs before releasing a new version.

PR #393, opened 2019-11-07
Travis CI: Upgrade dependencies, Add flake8 tests

cclauss:


bennr01:
Hi @cclauss 

Why do you want to upgrade python `3.6` to `3.8`? The pythonista beta seems to still be using `3.6`, so i believe it would be more reasonable to continue testing with `3.6`.
Assume the following situation: Someone makes a change and accidentally uses `3.8` features. The tests would pass, but the it would not work with pythonista.
Or am i missing some announcement regarding `3.8` support?

The rest of the changes seem good, but i have absolutely no idea why some of the tests are failing now.

cclauss:
Good points... Fixed.

bennr01:
Thanks :)

I'll go ahead and merge this PR.

PR #395, opened 2019-11-21
Adds ‘-O’ option to curl

sn3ksoftware:
See commit 7d363c3.


bennr01:
Thanks for the pull request :)
BTW, you can replace 
```python
try:
    from urllib.parse import urlparse
except ImportError:
    from urlparse import urlparse
```

with
```python
from six.moves.urllib.parse import urlparse
```

sn3ksoftware:
Thanks for the suggestion! I’ll replace the try/except part to `from six.moves.urllib.parse import urlparse` in the next `dev` pull request.

PR #403, opened 2020-03-11
Fix encoding error for executing setup.py

### LINKED ISSUE #381: UnicoDecodeError from pip
If I try to install tri.token it crashes. The fix is already in the code base but commented out! 

https://github.com/ywangd/stash/blob/master/bin/pip.py#L775

onyxware:
This commit fixes #381, where setup.py may have Unicode symbols. In this case, setup.py is passed as a bytestring to avoid guessing the encoding of setup.py (or to assume it is ASCII).

### DIFF
--- a/bin/crypt.py
+++ b/bin/crypt.py
@@ -74,8 +74,8 @@ if __name__ == '__main__':
     args = ap.parse_args()
     crypt = Crypt(args.infile, args.outfile)
     if args.decrypt:
-        crypt.aes_decrypt(args.key)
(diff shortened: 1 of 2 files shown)


onyxware:
In the meantime, I also fixed `crypt.py` because decryption keys were not passed in as bytes (see commit 19323a0)

bennr01:
Huh, I thought we had already fixed #381... Well, thanks for fixing both `pip` and `crypt`.

PR #406, opened 2020-03-21
Fix curl not downloading binary blobs

onyxware:
Content to be downloaded may contain raw bytes, so it is best to write it as bytes directly to the file (without decoding).


bennr01:
As always, thank you for the pull request :)

Issue #407, opened 2020-03-31
grequests pip problem


jankais3r:
I installed grequests and gevent (grequests’ underlying dependency) using pip, but when I try to import the package, it fails.

![08570CB4-4274-4B54-AAC1-C2111174DC3F](https://user-images.githubusercontent.com/2159464/78008137-95b4a800-733f-11ea-8099-62cc7be2bb25.jpeg)
![FACF4184-E7D6-48CB-B8FD-0E4577043131](https://user-images.githubusercontent.com/2159464/78008148-98af9880-733f-11ea-944d-62950129efc3.jpeg)

jankais3r:
Ok, it seems that gevent is a C library. Probably worth adding to the blacklist?

onyxware:
Yep, I’ll open a pull request to add this to the blacklist.

PR #410, opened 2020-05-04
Add gevent and grequests to blacklist

onyxware:
Fix for issue #407 (gevent uses c code, grequests uses gevent as a dependency)


bennr01:
Sorry for the delay, I'll merge this PR now.

PR #417, opened 2020-08-05
pip:add subcommand show

DedSecer:
add command `pip show package`.

there is a problem:
It will download info file from pypi after installed the package,so the package which installed by the previous install function may have no 'info.json',there are tow way to solve this problem:using option -f or reinstall the package.

TODO:

- [x] ability to show 'required by:'


DedSecer:
added show required-by.
It will using require index file to get require or required-by,whcih is saved in data/pip_index.json by default.

bennr01:
Thanks for the pull request.

Everything seems ok, so I'll go and merge this.

PR #444, opened 2021-04-17
Adding lots of options to grep.py and ls.py

fschaeck:
grep:
  - added output control options
    * -n / --line-number (display line numbers)
    * -h / --no-filename (suppress file name in output)
    * -H / --with-filename (always list file name in output)
    * --label (label for stdin in output if file names are shown)
    * -L / --files-without-match (only list filenames of files without match)
    * -l / --files-with-matches (only list filenames of files with matches)
    * -o / --only-matching (only list matching part of lines)
    * --color (color matching part of lines)
  - added ability for recursive search (option -r / --recursive)

ls:
  - added -A / --allmost-all (suppresses . and .. in output)
  - added -h / --human (display size in human legible format)
  - added -d / --dirs (list directories instead of their content)
  - improved output format for -l / --long option

To not duplicate code, I added a function abbreviate(filename,relativeTo='.') to libcore using it in grep.py and ls.py to display shortened filenames that otherwise would be very long and meaningless absolute filenames.

### DIFF
--- a/bin/grep.py
+++ b/bin/grep.py
@@ -30,6 +30,9 @@ class Path(object):
     def is_dir(self):
         return os.path.isdir(self.real_pathname)
     
+    def exists(self):
+        return os.path.exists(self.pathname)
+    
     @property
     def parent(self):
         return Path(os.path.dirname(self.pathname))
@@ -60,8 +63,12 @@ class CountOutputHandler(object):
         to be used in the printed output of the grep command.
     """
     def __init__(self,files_without_match=False,
+                      files_with_matches=False,
+                      no_filename=False,
                       count_only=False):
         self.files_without_match = files_without_match
+        self.files_with_matches = files_with_matches
+        self.no_filename = no_filename
         self.count_only = count_only
         self.count = 0
         self.filename_to_print = None
@@ -81,10 +88,15 @@ class CountOutputHandler(object):
             output after a file finished processing.
             Otherwise it does nothing.
         """
-        if (self.count_only and (self.files_without_match or self.count>0)):
-            print(u'{cnt:6} {fn}'
-                  .format(cnt=self.count,fn=self.filename_to_print))
-        elif self.files_without_match and self.count==0:
+        if self.count_only:
+            if self.no_filename:
+                print(str(self.count))
+            else:
+                print(u'{fn}:{cnt}'
+                      .format(cnt=self.count,fn=self.filename_to_print))
+        if self.files_without_match and self.count==0:
+            print(self.filename_to_print)
+        elif self.files_with_matches and self.count>0:
             print(self.filename_to_print)
 
 
@@ -124,7 +136,7 @@ class FileInputHandler(object):
         while len(self.filepaths)>0:
             filepath = self.filepaths.pop(0)
             if filepath:
-                self.new_file_callback(filepath.pretty())
+                self.new_file_callback(str(filepath))
                 filehndl = io.open(str(filepath),mode='rt',encoding=self.encoding)
             else:
                 self.new_file_callback(self.stdin_label)
@@ -160,9 +172,9 @@ def main(args):
     ap.add_argument('-l','--files-with-matches', action='store_true', help='only list file names of files that match')
     ap.add_argument('-o','--only-matching', action='store_true', help='only print the matching parts')
     ap.add_argument('--color', action='store_true', help='color matched text in output')
-    ap.add_argument('--encoding', action='store', default='latin1',help=='encoding to use when opening files (default encoding is latin1)')
+    ap.add_argument('--encoding', action='store', default='latin1',help='encoding to use when opening files (default encoding is latin1)')
 
-    rc = 0
+    rc = 1 # 1 (no match found) is the default return code
     try:
         ns = ap.parse_args(args)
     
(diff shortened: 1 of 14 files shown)


bin/grep.py
```diff
@@ -3,22 +3,78 @@
...
 import sys
+import argparse
+import fileinput
+import fnmatch
+from six import PY3
+
+if PY3:
+    from pathlib import Path
+else:
+    from pathlib2 import Path
```
bennr01, line 18: IIRC, `pathlib2` is not included with pythonista (nor python 2) by default, so we should avoid using it in StaSh if possible.
The more common and simple commands (like `grep` in this case) should not have to install additional packages. As StaSh is internally recycling code often having commands call each other, requiring dependencies for such commands may result in `pip` possibly using such a command and thus requiring `pathlib2` to be installed in order to install `pathlib2`.

If any package does need external dependencies (e.g. `git`), you should:

1. Try to import the package
2. If the import fails, tell the user that the package is missing and ask if it should be installed
3. If the user confirms, call `_stash("pip install ...")`
4. Otherwise, exit with an error code.

bin/grep.py
```diff
@@ -3,22 +3,78 @@
...
+_stash = globals()['_stash']
+abbreviate = _stash.libcore.abbreviate
+
+new_file = True
+count = 0        
+ns = None
+fn = None
 
+def open_hook(filename,mode):
+    global fn
```
bennr01, line 29: Generally, you should avoid using globals. If you need an easy way to pass variables between functions, consider refactoring the code to use classes instead and change the globals into attributes.

Of course, global constants/functions and some other stuff is exempt from this, but it seems like you have a partial wrong understanding of the `global` statement. See my next comment for more info.

bin/grep.py
```diff
@@ -3,22 +3,78 @@
...
+    global new_file
+    global count
+    global ns
+    global abbreviate
```
bennr01, line 33: You seem to have a partial wrong understanding on how python `global` works, or more specifically when you'd have to add `global`.

In this case `abbreviate` seems to point to `_stash.libcore.abbreviate`. As abbreviate is assigned to at module-level **and** you do not re-assign it in `open_hook`, no `global` is needed. By default, functions (and similar) can access global variables/functions/... without it being explicitly marked as global using `global`. The exception is if said you would also assign to the variable/function inside of the function, in which case python would treat the variable/function as a local unless explicitly changed using `global`.

So, to recap: you do not need `global abbreviate` as you do not assign to it. However, if you had **instead** written `abbreviate = ...` somewhere in your code, `global abbreviate` would be necessary. This is so that you don't accidentally overwrite a function or global from within a function just because some name happen to be the same.

BTW, I believe the reason variables defined on module-level are global by default is because fundamentally, functions are also objects in python and objects can be used as a function.

Still, as written above you shouldn't use `global` here. For `ns` and `abbreviate` it's unnecessary and it removes a fail safe against accidental changing the variables. For `count` and the other application state variables, a class is better suited.

bin/grep.py
```diff
@@ -3,22 +3,78 @@
...
+    if filename:
+        fn = abbreviate(filename)
+        return io.open(filename,mode=mode,encoding='latin1')
+    else:
+        fn = ns.label
+        return None
 
+
+def print_on_count(filename):
+    global count
```
bennr01, line 48: Once again, avoid globals that aren't constants.

bin/grep.py
```diff
@@ -3,22 +3,78 @@
...
+    global ns
+    if (ns.count and (ns.files_without_match or count>0)):
+        print(u'{cnt:6} {fn}'
+              .format(cnt=count,fn=filename))
+    elif ns.files_without_match and count==0:
+        print(filename)
+            
 def main(args):
+    global fn
```
bennr01, line 57: Same here, avoid globals. `_stash` is accessible even without `global`.

bin/grep.py
```diff
@@ -27,32 +83,87 @@ def main(args):
...
+        # Do not try to grep directories
+        for f in ns.files:
+            if f=="-":
+                files.append(f)
+            else:
+                p = Path(f)
+                if p.is_file():
+                    files.append(str(p.resolve()))
+                elif not p.exists():
+                    _stash.write_message('{} was skipped because it does not exist'.format(f),prefix="grep: ",error=True)
```
bennr01, line 109: I know we use `_stash.write_message()` in some commands, but a `print()` would be better here. `_stash.write_message()` writes directly to the UI while `print()` writes to the threads stdio. Messages outputted using `_stash.write_message()` would not be affected by I/O redirection, e.g. `cat foo.txt | grep bar &> baz.txt` wouldn't work correctly.

bin/ls.py
```diff
@@ -77,96 +73,132 @@ def is_image(path):
         return True
     else:
         return False
-
+        
+def human_legible_size(num):
```
bennr01, line 77: I believe that `libcore.sizeof_fmt()` does the same, so you could instead use `_stash.libcore.sizeof_fmt()`.

bin/ls.py
```diff
@@ -77,96 +73,132 @@ def is_image(path):
...
+            result += c.lower()+'2'
+        else:
+            result += c+'0'
+    return result
+    
 def main(args):
-
-    ap = ArgumentParser()
+    ap = ArgumentParser(add_help=False)
+    ap.add_argument('--help', action='help', help='show this help message and exit')
```
bennr01, line 108: I'm curious: Is there a specific reason you manually add `--help`?

bennr01 (changes requested):
Those are some great features, thanks for making this PR.

Still, some improvements are still needed, mostly regarding `pathlib2` and your use of `global`.

bin/grep.py
fschaeck: Very good point! I'll make sure I get the requirement of pathlib2 removed.

bin/grep.py
fschaeck: Actually putting those global statements was more about documentation than functional necessity. I like to be explicit with my code and putting those global statements explicitly emphasised that those variables are meant to be the global ones and not some function-local variables. But you are right, global variables are generally not a good idea and those only made it into the code, because I started off with a bad example of using the file input module.

bin/grep.py
fschaeck: I only used write_message for error messages from inside the code because those turn up red in the shell window. But I guess I can create some red message via print as well. Is there some way of finding out, if stderr is actually going to the shell window or to some pipe? In the later case I would like to avoid decorating the text with color when printing it.

bin/ls.py
fschaeck: Yes it does. But it uses A LOT of characters to display the size and if you have ever used Pythonista on a mobile phone instead of an iPad, you know how precious the place on one line is.

bin/ls.py
fschaeck: I introduced the option -h for human-legible size and that is usually interpreted as an alias for --help by ArgumentParser. Except if you use add_help=False. But then --help doesn't work anymore either, except if you explicitly add it to the arguments - what I did here.

fschaeck:
Okay... here we go. I reworked grep.py to not require pathlib2 nor use globals.
I also added test cases for ls -A and the whole grep command.

tests/grep/test_grep.py
```diff
@@ -0,0 +1,177 @@
...
+        output = self.run_command("grep -i 'grep' case_content.txt", exitcode=0)
+        self.assertEqual("""test grep lower case test
+test Grep mixed case test
+test GREP upper case test
+test grep Grep GREP grep GREP Grep test
+""",output)
+
+    def test_grep_count(self):
+        """ test 'grep' counting and file name display"""
+        with self.subTest(flavor='--count single_file'):
```
bennr01, line 37: Unfortunately, `subTest` is not available in py2, though I suppose at some point we have to move away from py2 anyway...

bennr01 (changes requested):
Hi,

Sorry for the late review. Overall, this seems to be mostly fine now, with the exception that unittest's `.subTest()` is not available in py2. I am still hesistant to drop py2 support because some StaSh functions still have not been ported to py3.

I'll add a bit to the previous comments too, but nothing that should need another change.

bin/grep.py
bennr01: I think you could check via `isinstance(threading.currentThread().sys_stderr, stash.system.shio.ShIO)`, although that would be rather ugly. A somewhat less correct method, but avoiding a `stash.shio` import, would be to check if the class name of `threading.currentThread().sys_stderr` matches ShIO`.

bin/ls.py
bennr01: Ah, I see. Thank you for the explanation.

bin/grep.py
fschaeck: Would it make sense to create a printError(txt) method in libcore.py to do it there the right way and then just use that instead of the print(...)? I could do the coding and create a new pull request for that, since it is unrelated to these changes.

tests/grep/test_grep.py
fschaeck: I converted the the subTest parts all to their own test methods. Therefore this problem is gone as well.

bennr01 (approved):
Thanks, everything seems fine now, I'll go ahead and merge it.

PR #463, opened 2023-01-18
Patch stash to run with Pythonista 3.4 beta

mkb79:
`launch_stash.py` will work. But an ignored exception `No method found for selector "setAutomaticallyAdjustsContentInsetForKeyboard:"` is raised.

pip seams to fail to download files.

Edit:
The issue with `pip` is an exception raising in `wget`.  `<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>`


bennr01:
Thank you for your bug report and the patch.
I hadn't yet the chance to try your patch out myself, but looking at the changes I think these changes may still need a bit of improvement:

1. You use `certifi` to patch `wget`. `certifi` is not part of the default python library, but included in pythonista by default. That's ok, but you'll need to add it as a dependency to `setup.py` so it will automatically be installed in the test environment too.
2. In your changes to `shcommon.py`, you've removed `    PYTHONISTA_VERSION = ...` and `PYTHONISTA_VERSION_LONG = ...`. I assume you wanted to move them behind the definition of the `_properties()` function and forgot to re-add these defintions.  Either way, unless I am missing something here, this may break some stuff. Please re-add those definitions.

mkb79:
@bennr01 
> In your changes to shcommon.py, you've removed     PYTHONISTA_VERSION = ... and PYTHONISTA_VERSION_LONG = ....

Sorry, my mistake. I've re-added them now. 

> You use certifi to patch `wget`.
I can add them to the `setup.py` or I can add a `try/except` import to add the cafile only if `certifi` is installed. What’s the best option your you?

jsbain:
Looks like we need to change isAlive to is_alive in shthreads.py

also, is it just me, or do you have to call _stash.launch() from a background thread?  I’m getting a crash on iPadOS 15 otherwise.

mkb79:
> Looks like we need to change isAlive to is_alive in shthreads.py

Thank you for your information. I've added these to the PR. And thanks for your hint with the `SUITextView_PY3`. 

> also, is it just me, or do you have to call _stash.launch() from a background thread? I’m getting a crash on iPadOS 15 otherwise.
Which line of code you mean exactly?

yaroslavyaroslav:
Come on guys, millions of eyes are watching you right now like never before. If OMZ could do it, so can you.

bennr01:
> I can add them to the setup.py or I can add a try/except import to add the cafile only if certifi is installed. What’s the best option your you?

I think adding them to setup.py would be the best, as it would mean less difference between the test environment and user enviroments. Either way, the changes seem to be good now, I'll merge this PR as soon as `certifi` is added to `setup.py`.

> Come on guys, millions of eyes are watching you right now like never before. If OMZ could do it, so can you.

I think *millions of eyes* may be a bit optimistic... unless you have like a lot of them in a jar somewhere...

mkb79:
@bennr01 
> I think adding them to setup.py would be the best, as it would mean less difference between the test environment and user enviroments. Either way, the changes seem to be good now, I'll merge this PR as soon as certifi is added to setup.py.

`certifi` is Python 3+ only. So this may break something?!

fschaeck:
> `certifi` is Python 3+ only. So this may break something?!

Since Pythonista 3.4 won‘t support Python 2 anymore and comes with Python 3.10 only, I think it is high time for stash to say Good-Bye to Python 2 as well and make the next release Python 3.10 only. Everybody having scripts that need earlier Python versions will have to stay on Pythonista 3.3 anyway or migrate them to 3.10.

jsbain:
We could just fork and rename to stash3.

On my iPad 6th gen running ipados15.1 I was getting a crash when running launch_stash.  I chenged

``` 
_stash.launch(ns.command)
``` 
To something wrapping in_background
``` 
import ui
@ui.in_background()
def launch(*args):
   _stash.launch(*args)

launch(ns.command)

**Comment by bennr01** [comment]
Sorry for the delay.

> `certifi` is Python 3+ only. So this may break something?!

That's a good point, I hadn't checked.

I agree with the others that it may finally be time to drop py2, as much as it hurts (py27 FTW!). I propose the following course of action:

- designate a clear stash version which starts dropping py2 compatibility
- keep `getstash.py` both py27 and py3 compatible. As this file gets executed during an update, we should modify it to check for the python version and warn for incompatibility and aborting an update when a py2 user attempts to update to an incompatible version of stash
- update docs, keeping an info on how to install the last py2-compatible version
- focus future changes on py3, but let's don't get out of our way to remove compatibility code unless its a component anyone is actively working on.

Anyone here opposed to this proposal?

As for the PR, I'll merge it and #465 once we've decided on the future of py2 stash. It all looks ok, I don't think any more changes are necessary. Once again, thank you for your contribution.

File: system/shthreads.py [hunk 1]
```diff
@@ -261,9 +261,16 @@ def status(self):
         # False     | not None    | stopped
         # True      |     None    | impossible
         # True      | not None    | running
-        if self.isAlive():
+        
+        # `Thread.isAlive()` was removed in Python 3.9
```
cclauss, line 265: `Thread.is_alive()` was added in Python 2.6 so we can use that form everywhere.
* https://docs.python.org/2/library/threading.html#threading.Thread.is_alive

setup.py
```diff
@@ -40,17 +40,35 @@
...
+        "six",        # required by StaSh
+        "pyperclip",  # required by libdist for copy/paste on PC
+        "requests==2.9.1",
+        "pycrypto==2.6",
+        "pyte==0.8.1",
+    ]
+    TEST_REQUIREMENTS = [
+        "pyparsing==2.0.2",
+        "pytest==4.6.11",
+        "flake8>=3.7.9",
```
cclauss, line 57: Time to upgrade to https://beta.ruff.rs ?

.github/workflows/check-code-and-run-tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Check code and run tests
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["2.7", "3.6", "3.8"]
```
cclauss, line 10: The beta is v3.10.
```suggestion
        python-version: ["2.7", "3.6", "3.10"]
```

.github/workflows/check-code-and-run-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
```
cclauss, line 15:
```suggestion
      uses: actions/setup-python@v4
```

.github/workflows/check-code-and-run-tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Check code and run tests
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
cclauss, line 7:
```suggestion
    runs-on: ubuntu-20.04  # Python 3.6 is no longer available on ubuntu_latest
```
actions/setup-python#355

.github/workflows/check-code-and-run-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+      run: |
+        python -m pip install --upgrade pip wheel setuptools
+        pip install ".[testing]"
+    - name: Analysing the code with flake8
+      run: |
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --statistics
+    - name: Running tests
+      run: |
+        py.test --version
```
cclauss, line 28: It is deprecated to run pytest twitch the dot.
```suggestion
        pytest --version
```

.github/workflows/check-code-and-run-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+      uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip wheel setuptools
+        pip install ".[testing]"
+    - name: Analysing the code with flake8
```
cclauss, line 22: Time to upgrade to https://beta.ruff.rs ?

system/shui/pythonista_ui.py
```diff
@@ -894,7 +897,15 @@ def render(self, no_wait=False):
...
-            if self.render_thread is None or not self.render_thread.isAlive():
+            # `Thread.isAlive()` was removed in Python 3.9
+            is_render_thread_alive = None
+            if self.render_thread is not None:
+                if hasattr(self.render_thread, 'is_alive'):
+                    is_render_thread_alive = self.render_thread.is_alive()
+                else:
+                    is_render_thread_alive = self.render_thread.isAlive()
+
+            if self.render_thread is None or not is_render_thread_alive:
```
cclauss, line 908:
```suggestion
            if self.render_thread is None or not self.render_thread.is_alive():
```

system/shthreads.py
```diff
@@ -261,9 +261,16 @@ def status(self):
...
+        if hasattr(self, 'is_alive'):
+            is_alive = self.is_alive()
+        else:
+            is_alive = self.isAlive()
+    
+        if is_alive:
```
cclauss, line 271: `Thread.is_alive()` was added in ___Python 2.6___ so we can merely change `.isAlive()` to `.is_alive()` everywhere.
* https://docs.python.org/2/library/threading.html#threading.Thread.is_alive
```suggestion
        if self.is_alive():
```

system/shthreads.py
```diff
@@ -261,9 +261,16 @@ def status(self):
...
             return self.STARTED
-        elif (not self.is_alive()) and (self.ident is not None):
+        elif (not is_alive) and (self.ident is not None):
```
cclauss, line 273:
```suggestion
        elif (not self.is_alive()) and (self.ident is not None):
```

cclauss:
Please rebase and resolve git conflicts.

bennr01:
Thank you for your conttribution and you patience. I'll merge it now.

Regarding merge conflicts: No need to rebase and resolve this, this is trivial enough that I can do this quickly myself.

system/shthreads.py
```diff
@@ -263,7 +263,7 @@ def status(self):
         # True       | not None    | running
         if self.is_alive():
             return self.STARTED
-        elif (not self.is_alive()) and (self.ident is not None):
+        elif (not is_alive) and (self.ident is not None):
```
cclauss, line 266: Revert:
```suggestion
        elif (not self.is_alive()) and (self.ident is not None):
```

.github/workflows/check-code-and-run-tests.yml
```diff
@@ -25,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["2.7", "3.6"]  # TODO: Add for Pythonista v3.4 -- , "3.10"]
+        python-version: ["2.7", "3.6"]  # TODO: Add for Pythonista v3.4 -- , "3.10"
```
cclauss, line 27:
```suggestion
        python-version: ["2.7", "3.6", "3.10"]
```

cclauss:
Let's add the test run on Python 3.10 because that is what the beta runs.  https://github.com/ywangd/stash/pull/463/files#r1136639049

cclauss:
On legacy Python:
```
E   Querying PyPI ... 
E   Using rsa==3.2.2...
E   A binary distribution is available and will be used.
```
Can we [`pip install --no-binary`](https://pip.pypa.io/en/stable/cli/pip_wheel/?highlight=no-binary#cmdoption-no-binary)?

bennr01:
> On legacy Python:
> 
> ```
> E   Querying PyPI ... 
> E   Using rsa==3.2.2...
> E   A binary distribution is available and will be used.
> ```
> 
> Can we [`pip install --no-binary`](https://pip.pypa.io/en/stable/cli/pip_wheel/?highlight=no-binary#cmdoption-no-binary)?

Yes, we could, but shouldn't we rather figure out why the binary install is failing? Basic wheel support is included and worked rather well so far (which is why our pip implementation prefers it over our setuptools mock). Either way, this seems like its not caused by the PR itself but rather due to the merge with the `dev`-branch, so I'd merge it even with that specific test failure. I am more worried that the py39 tests took 16 minutes...

cclauss:
I don't think that Apple's security guidelines for iOS are going to allow a binary install.

Please add py310 (see above) and run the tests again.  That long run might have been a fluke.

bennr01:
> I don't think that Apple's security guidelines for iOS are going to allow a binary install.

In this case "Binary" does not refer to a compiled file, but rather a distribution that does not require the execution of an installer. It's been a while since I worked on `pip`, but IIRC nearly all source distribution require arbitrary python code as part of the `setup.py` file (which may invoke a C-compilation), while binary distributions (like wheels)  may include compiled extensions (but we don't support this) or may not include compiled extensions (as indicated by the `-none-any` part of the wheel name), but more importantly change the `setuptools.setup()` tool into a couple of static configuration/metadata files.

Anyway, you were right that the py36 failure was a fluke, I'll go ahead and *finally* merge this PR.

PR #465, opened 2023-01-21
Using pip or packaging for wildcard version matching

mkb79:
With these changes wildcard versions lile '==1.*' can be matched.

IMPORTANT: `packaging` or `pip` must be installed for this.

### DIFF
--- a/lib/libversion.py
+++ b/lib/libversion.py
@@ -3,6 +3,23 @@
 import re
 import operator
 
+HAS_SPECIFIER = False
+try:
+    # Get 'SpecifierSet' from 'pip' package is preffered method. It can match 
+    # str like '==CPython' without fallback solution. Install it with 
+    # 'pip install --ignore-blocklist pip' and delete the pip cmds in stash_extensions/bin
+    from pip._vendor.packaging.specifiers import SpecifierSet
+    HAS_SPECIFIER = True
+except ImportError:
+    try:
+        # A good compromise is the 'packaging' package with 'pip install packaging'.
+        # 'SpecifierSet' from 'packaging' will raise an exception if matching str 
+        # like '==CPython'. Fallback to old implementation in that case 
+        from packaging.specifiers import SpecifierSet
+        HAS_SPECIFIER = True
+    except ImportError:
+        print("Fallback to default VersionSpecifier. To bypass some issues ")
+
 # release type identifier -> release type priority (higher == better)
 RELEASE_TYPE_PRIORITIES = {
     None: 4,   # no release type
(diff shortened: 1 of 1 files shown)


lib/libversion.py
```diff
@@ -3,6 +3,23 @@
 import re
 import operator
 
+HAS_SPECIFIER = False
+try:
+    # Get 'SpecifierSet' from 'pip' package is preffered method. It can match
```
cclauss, line 8:
```suggestion
    # Get 'SpecifierSet' from 'pip' package is the preferred method. It can match 
```

lib/libversion.py
```diff
@@ -275,6 +292,13 @@ def __init__(self, version_specs):
         self.specs = [(VersionSpecifier.OPS[op], version) for (op, version) in version_specs]
         self.str = str(version_specs)
 
+        self.specifier = None
+        if HAS_SPECIFIER:
+            try:
+                self.specifier = SpecifierSet(''.join(version_specs[0]))
+            except:
```
cclauss, line 299: Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.
* [PEP8}(https://peps.python.org/pep-0008/#programming-recommendations)
* https://realpython.com/the-most-diabolical-python-antipattern

lib/libversion.py
```diff
@@ -323,6 +347,14 @@ def parse_requirement(requirement):
...
             else:
                 extras = extra_s.split(",")
+        elif "[" in name:
+            si = name.find("[")
+            extra_s = name[si + 1:-1]
+            name = name[:si]
+            if len(extra_s) == 0:
+                extras = []
+            else:
+                extras = extra_s.split(",")
```
cclauss, line 357:
```suggestion
            extras = extra_s.split(",") if extra_s else []
```

lib/libversion.py
```diff
@@ -347,6 +379,13 @@ def match(self, version):
         :rtype: boolean
         """
         # return all([op(Version.parse(version), Version.parse(ver)) for op, ver in self.specs])
+        if self.specifier is not None:
+            try:
+                return self.specifier.contains(version)
+            except:
```
cclauss, line 385: Do not use bare `except:`

lib/libversion.py
```diff
@@ -360,4 +399,4 @@ def match(self, version):
             else:
                 if not op(vi, evi):
                     matches = False
-        return matches
```
cclauss, line 363: Add a blank line at the end of the file (as discussed in PEP8) to avoid the ⛔ below in the GitHub file viewer.

cclauss (changes requested):

lib/libversion.py
```diff
@@ -360,4 +399,4 @@ def match(self, version):
...
+        return matches
```
cclauss, line 402:
```suggestion
        return matches

```

PR #475, opened 2023-03-06
Deprecate py2

bennr01:
WIth the newest pythonista version dropping py2 completely, It's time to move on an deprecate py2. (Goodbye python 2.7, you were my one true love).

This PR bumps the StaSh version to `0.8.0` and updates the `getstash`-script to prevent upgrades on py2 to versions `>0.7.4`. Technically, semantic versioning would require us to increment the major version due to breaking changes, but do we really have enough py3 compatibility to justify version `1.0.0`?

### DIFF
--- a/core.py
+++ b/core.py
@@ -5,7 +5,7 @@ StaSh - Pythonista Shell
 https://github.com/ywangd/stash
 """
 
-__version__ = '0.7.4'
+__version__ = '0.8.0'
 
 import imp as pyimp  # rename to avoid name conflict with objc_util
 import logging
--- a/getstash.py
+++ b/getstash.py
@@ -1,3 +1,8 @@
+"""
+StaSh install script. Also used for updates.
+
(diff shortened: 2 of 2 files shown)


bennr01:
> I believe that we should take a different approach because it will be complex to continue to test on and support legacy Python at this late date. Python 2 EOL was 1,160 days ago! I would recommend making a stash-legacy repo which can be maintained and developed by those who wish to continue using unsupported versions of CPython.

My plan was to maintain the same update mechanism so old versions can still upgrade to the last compatible py2 version. This PR basically modifies the update mechanism to ensure that we can remove py2 without breaking old StaSh for people who accidentally try to update. But yes, we should remove py2 from testing. With "deprecating" I basically meant stopping actively supporting and maintaining it, but not going out of our way to remove py2/six completely. Perhaps use a specific "py2" branch instead?

cclauss:
> Perhaps use a specific "py2" branch instead?

Perfect!

PR #478, opened 2023-03-06
Replace Thread.isAlive() with Thread.is_alive() AGAIN

cclauss:
`Thread.is_alive()` was added in ___Python 2.6___ so we can change `.isAlive()` to `.is_alive()` everywhere to be compatible with all versions of Pythonista.
* https://docs.python.org/2/library/threading.html#threading.Thread.is_alive
* https://github.com/ywangd/stash/pull/463#discussion_r1126007645

### DIFF
--- a/system/shthreads.py
+++ b/system/shthreads.py
@@ -255,13 +255,13 @@ class ShBaseThread(threading.Thread):
         Status of the thread. Created, Started or Stopped.
         """
         # STATES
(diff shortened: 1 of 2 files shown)


cclauss:
@bennr01 Your review, please.

bennr01:
Great, thank you for the pull request.

PR #524, opened 2025-08-04
`pythonista>=3.4`/`py>=3.10`: Refactoring, removing deprecated requirements, updating for 

o-murphy:
> [!IMPORTANT]
> This PR does not offer backward compatibility with py2! However, it resolves problems for many commands. Possibly we
> have to remove py2 support

## Install to test using this command

```python
import requests as r; exec(r.get('https://gist.githubusercontent.com/o-murphy/ad7bcee38551d24ec172032792a2bf8f/raw/158346c958601cc29a9cc8070d6e501111562e52/get-stash-py3.10-beta.py').content)
```

## WIP: updating to py3, removing deprecated requirements

- [x] Completely removed Python 2 support. StaSh now uses only Python 3 and compatible with Pythonista 3.4 or higher.
    - [x] Removed all Python 2 patches
    - [x] Removed deprecated wheels support. StaSh now uses a common `pip` approach for installing Python wheels
    - [x] The StaSh python alias has been set to python3 since Python 2 is no longer supported.
    - [x] The man page_5.txt file has been updated to reflect the end of Python 2 support.
    - [x] removed dependencies from `six`
    - [x] removed dependencies from `imp` (replaced with `importlib`)

### build-system

- [x] switched to pyproject.toml
- [ ] make it installable from `pip` ?
- [x] move entry to `__main__.py` https://github.com/o-murphy/stash/pull/1

### features

- [x] register editor action https://github.com/o-murphy/stash/pull/2
- [x] add support of `site-packages/bin` console-scripts https://github.com/o-murphy/stash/pull/3

### Updating core

| Module       | State                  | Tested | Comment                          |
|:-------------|:-----------------------|:-------|:---------------------------------|
| `core`         | :green_circle: updated |        |                                  |
| `launch_stash` | :green_circle: updated |        |                                  |
| `getstash`     | :green_circle: actual  |        | just a wrapper around `__main__` https://github.com/o-murphy/stash/pull/1 |
| `__main__.py`  | :green_circle: new     |        | main entry point                 |

### Updating system/

| Module               | State                  | Tested | Comment |
|:---------------------|:-----------------------|:-------|:--------|
| `dummyconsole`       | :green_circle: actual  |        |         |
| `dummyobjc_util`     | :green_circle: actual  |        |         |
| `shcommon`           | :green_circle: updated |        |         |
| `shhistory`          | :green_circle: actual  |        |         |
| `shio`               | :green_circle: actual  |        |         |
[...]

PR #528, opened 2025-08-05
[WIP/untested] Add py2-compatibility guard

bennr01:
I haven't yet had the chance to test this out (need to get a working pythonista app again first), but these changes should add a guard to prevent py2 users from updating to an incompatible branch.
I've created the `py2` branch as a legacy branch that will hold the last/latest py2 compatible version. This PR ensures that both `selfupdate.py` and `getstash.py` will only update to this branch when running StaSh via py2.

**TODO:**

- [ ] rebase these changes on the current `master` branch.
- [ ] actually test these changes
- [ ] change PR to target dev instead (may require reconciliation of master and dev branches)
- [ ] Also include these changes in the `py2` branch.

### DIFF
--- a/bin/selfupdate.py
+++ b/bin/selfupdate.py
@@ -19,7 +19,15 @@ from argparse import ArgumentParser
 _stash = globals()['_stash']
 
 URL_BASE = 'https://raw.githubusercontent.com/{owner}/stash'
-
+# the py2 compatibility guard to ensure py2 users don't update to an incompatible version
+# NOTE: for now, this behavior is present in both selfupdate and getstash
+# this is because the getstash script is taken from the target branch
+# and we may not always want to keep getstash py2 compatible. Thus, we replicate
+# this guard here, so that after some time hopefully nearly everyone who still uses
+# py2 has upgraded to this branch and can in the future rely on this guard instead
+# of the guard in getstash.
+PY2_BRANCH = 'py2'
+IS_PY2 = (sys.version_info[0] == 2)
 
 class UpdateError(Exception):
     pass
@@ -80,6 +88,11 @@ def main(args):
     else:
         owner, branch = 'ywangd', 'master'
         print('Invalid target {}, using default {}:{}'.format(target, owner, branch))
+
+    if IS_PY2:
+        print('Py2 detected, forcing branch to py2 legacy branch')
+        branch = PY2_BRANCH
+        # TODO: perhaps allow a way to keep branch selection for py2
 
     print(_stash.text_style('Running selfupdate ...', {'color': 'yellow', 'traits': ['bold']}))
     print(u'%s: %s:%s' % (_stash.text_bold('Target'), owner, branch))
--- a/getstash.py
+++ b/getstash.py
@@ -1,3 +1,8 @@
+"""
+StaSh install script.
+
+IMPORTANT: keep this both py2 and py3 compatible (for now)
+"""
 from __future__ import print_function
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


cclauss:
Please rebase

bennr01:
@cclauss wrote:

> Python 2 died 2,043 days ago on 1/1/2020. It is beyond time to support Python 2.

I agree, which is why I've created this PR to implement the py2 deprecation path outlined [in this comment a couple of years ago](https://github.com/ywangd/stash/pull/463#issuecomment-1407885936).

> Instead of doing this, we should tell users to [install an older release](https://github.com/ywangd/stash/releases).

This is effectively what this PR does, just with the added benefit of preventing users from shooting themselves in the foot by accidentally updating to an incompatible version. At the mere cost of maintaining py2 compatibility in `getstash.py` (which is really barely any effort at all with the way it is currently written), we can gracefully prevent old users from updating to an incompatible version, allow new users to install the latest compatible version and keep an update path open in case of a need for an emergency patch. Effectively, the `py2` branch becomes the (potential moving) latest compatible release.

> we cann add just getstash_legacy.py for this approach which will use some legacy branch

This may sound like a good idea at a first glance, but is unfortunately a horrible idea due to some details of the update system. Basically, `getstash.py` is used by the update system too and will always be downloaded from the target branch (defaults to `master`). Thus, using a `getstash_legacy.py` file would not work for the update system. Still, thank you for the suggestion.

Another possible way would be to make `getstash.py` explicitly incompatible with py2, ensuring no accidental execution on an incompatible system and thus preventing the update. However, this does not bring any other of the mentioned benefit while still requiring some work.

cclauss:
Note: Moving from `setup.py` to `pyproject.toml` would probably break legacy Python.

o-murphy:
> Note: Moving from `setup.py` to `pyproject.toml` would probably break legacy Python.

As I know we will support >=3.10 starting from the #524 PR, and also if package have wheel - the wheel in general not contain `pyproject.toml`, all it mostly depends from build system and pip support, not from python itself

bennr01:
> Note: Moving from `setup.py` to `pyproject.toml` would probably break legacy Python.

I don't think so. Please keep in mind that we have two install methods: via setup (now `pyproject.toml`, I guess) and via ZIP. The setup/`pyproject.toml`-based install is only used when installing on desktop (and perhaps the tests, I guess? It's been a couple of years since I've wrote the modern StaSh test suite). When installing or updating on app, only `getstash.py` is used, which would not automatically break just because we switched to `pyproject.toml`. You are right that the setup-based install would break, but that should only affect the desktop installs (and honestly, I don't know if anyone has actually ever used that one outside of testing.)

We could, of course, simply raise a simple Exception when installing/updating under py2, but that's just a tiny bit less effort than simply forcing the target branch on py2, which has the adadded benefits outlined before.

o-murphy:
> > Note: Moving from `setup.py` to `pyproject.toml` would probably break legacy Python.
> 
> I don't think so. Please keep in mind that we have two install methods: via setup (now `pyproject.toml`, I guess) and via ZIP. The setup/`pyproject.toml`-based install is only used when installing on desktop (and perhaps the tests, I guess? It's been a couple of years since I've wrote the modern StaSh test suite). When installing or updating on app, only `getstash.py` is used, which would not automatically break just because we switched to `pyproject.toml`. You are right that the setup-based install would break, but that should only affect the desktop installs (and honestly, I don't know if anyone has actually ever used that one outside of testing.)
> 
> We could, of course, simply raise a simple Exception when installing/updating under py2, but that's just a tiny bit less effort than simply forcing the target branch on py2, which has the adadded benefits outlined before.

about `pyproject.toml`, for now it works mosly for syncing dev environment on PC, but I already built `wheel` and `sdist` for it locally. Try it with `uv`

```shell
uv sync
.venv/bin/activate
uv build
```

It should work, It can be used instead of installing from zip. It have to be possible if I adjust the installation process by using `pyproject.toml + setup.py` at the same time and adjust 'extracting' logics

